PR #1, opened 2020-08-30
README: add logo image

tigercosmos:
Looks like
image


jserv:
Thank @tigercosmos for contributing!

PR #2, opened 2020-09-02
Adjust maximum value in defs.h

eecheng87:
Change to smaller value to avoid memory exhaustion


jserv:

PR #5, opened 2020-09-06
Edit comment letter case to fit consistency

Mega-Pan:


jserv:
It was intended to use uppercase there.

PR #6, opened 2020-09-07
fix strcmp's logic for return values

reborn2266:
original implementation didn't follow the rule of classic
strcmp.


jserv:
Thank @reborn2266 for contributing!

PR #9, opened 2020-10-11
Support division operation

eecheng87:
Also un-comment some test bench in driver.sh


jserv:
Not all ARMv7-A core supports sdiv and udiv instructions. See https://patchwork.kernel.org/patch/1494331/
The CPU models such as cortex-a8 and cortex-a9 lack of sdiv/udiv instructions. You should mention this at least within git commit messages. Also, you can send another pull request to enforce `cortex-a7` support by modifying file `mk/arm.mk` with `-cpu cortex-a7`. In addition, the top-level `README.md` should be updated accordingly.

eecheng87:
> each ARMv7-A core supports sdiv and udiv instructions

commit message update!

jserv (changes requested):
Change the commit message to "SDIV might be unimplemented in some ARMv7-A
cores"

jserv:
Should be "Division operation" rather than "divide operation"

src/codegen.c
```diff
@@ -130,6 +130,7 @@ int get_code_length(ir_instr_t *ii)
     case OP_add:
     case OP_sub:
     case OP_mul:
+    case OP_sdiv:
```
jserv, line 133: Any strong reason to specify "sdiv" instead of "div" in general? We would like to support RISC-V later.

eecheng87: I suppose udiv will be imported later. If not possible, I can change it!

jserv:

src/codegen.c
lazyparser: > Any strong reason to specify "sdiv" instead of "div" in general? We would like to support RISC-V later.

Hi @jserv 

any plan/roadmap to support RISC-V?
I am planning to use shecc in my compiler (online) class and just start to replace the arm backend to RISC-V.
It would be a good news if you are implementing RISC-V backend.
(drive by comment. sorry for thread hijacking)

Issue #12, opened 2020-10-13
Fail to write global variable


eecheng87:
Hi, I am trying to add feature: global variable initialization.
I modify some code and it seems good when I dump IR, following is simple sample
```
...
0x000122d4         x1 := 44
0x000122d8         x0 = &a
0x000122e0         *x0 = x1 (4)
0x000122e4     main:
0x000122fc         {
0x000122fc             x0 := &data[36]
0x00012304             x1 = &a
0x0001230c             x1 = *x1 (4)
0x00012310             x0 := printf() @ 978
0x00012314             x0 := 0
0x00012318             return (from main)
0x0001231c         }
0x0001231c         exit main
```
corresponding C code

```cpp
int a = 44;
int main(int argc, char *argv[])
{
    printf("%d\n", a);
    return 0;
}
```

As I mention, it seems good for assigning initial value to `a` in IR, but the value printed is 0. This weird stuff stuck me a few day. Anything I ignore about writing value to global variable? Thanks!

jserv:
It is a known issue that global variable initialization is *not* supported. Therefore, you can not initialize a global such as `int i = [expr]`. In order to resolve this issue, we have to implement a pointer to bss/data section in code generation.

PR #13, opened 2020-10-14
Support global variable initialization

eecheng87:
In latest version, there already have .data section which is for storing **initialized global variable**. As a result, I don't add .bss section which is for uninitialized variable in this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,8 +165,7 @@ int fib(int n)      fib:                        Reserve stack frame for function
 
 1. Any non-zero value is NOT treated as logical truth by all ops.
    That is, the expression `0 == strcmp(ptr, "hello")` is not equivalent to `!strcmp(ptr, "hello")`.
-2. Global variable initialization is not supported. 
-   Therefore, you can not initialize a global such as `int i = [expr]`.
+2. The generated ELF lacks of .bss and .rodata section
 3. Dereference is incomplete. Consider `int x = 5; int *ptr = &x;` and it is forbidden to use `*ptr`.
    However, it is valid to use `ptr[0]`, which behaves the same of `*ptr`.
 4. The support of varying number of function arguments is incomplete. No `<stdarg.h>` can be used.
--- a/src/cfront.c
+++ b/src/cfront.c
@@ -548,6 +548,7 @@ int read_numeric_constant(char buffer[])
 
 void read_inner_var_decl(var_t *vd)
 {
+    vd->init_val = 0;
     if (lex_accept(T_asterisk))
         vd->is_ptr = 1;
     else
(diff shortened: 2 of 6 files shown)


jserv (changes requested):
Append "Close #12" in git commit message(s). Therefore, you can link a pull request to an issue to show that a fix is in progress and to automatically close the issue when the pull request is merged.
See https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

jserv:
You can remove the statement "2. Global variable initialization is not supported." in `README.md`.

tests/global.c
```diff
@@ -0,0 +1,13 @@
+int a = 44;
```
jserv, line 1: How about dropping `tests/global.c` and adding the original test in file `tests/driver.sh`?
I am planning to delete some trivial files in directory `tests` such as `def.c`, `fib.c`, `loop.c`, and `pointer.c`, which are already covered in test driver.

jserv: However, it still makes sense to generate an executable from `tests/hello.c`.

jserv:
You can prepare another git commit which applies this change to simplify the handing in file `src/globals.c`. The reason why function `global_init` exists is because the global variable initializations were not supported.

src/cfront.c
```diff
@@ -1308,6 +1309,25 @@ int read_body_assignment(char *token, block_t *parent)
...
 }
 
+int read_global_assignment(char *token)
+{
+    var_t *var = find_global_var(token);
+    if (var) {
+        char buffer[10];
+        /* global initialization must be constant */
+        /* TODO: support rvalue as expression e.g. int a = 3 + 2; */
+        if (lex_peek(T_numeric, buffer))
```
jserv, line 1319: Can you reuse the code in function `read_body_assignment`? That is, split some logic/operations into a helper function and then invoke the helper for reading the expressions from assignments.

jserv:
> I have changed commit with key word **Close #12** but nothing change. Did it work after PR merge?

If you check the end of #12, the issue is already associated with this pull request. Then, when this pull request is merged, #12 will be closed automatically.

src/cfront.c
eecheng87: Do I need to do that anyway? The reason I don't like to do that is:
* The only duplicate code in `read_body_assignment` is `find_global_var`
* Future update about global initialization will make more different with `read_body_assignment`. If I merge now, I guess `read_body_assignment` will become harder to read
* Merge control flow into `read_body_assignment` for one line re-use makes `read_body_assignment` more complicated

tests/global.c
eecheng87: Done! Move global initialization into `tests/driver.sh`

README.md
```diff
@@ -165,19 +165,18 @@ int fib(int n)      fib:                        Reserve stack frame for function
 
 1. Any non-zero value is NOT treated as logical truth by all ops.
    That is, the expression `0 == strcmp(ptr, "hello")` is not equivalent to `!strcmp(ptr, "hello")`.
-2. Global variable initialization is not supported. 
-   Therefore, you can not initialize a global such as `int i = [expr]`.
-3. Dereference is incomplete. Consider `int x = 5; int *ptr = &x;` and it is forbidden to use `*ptr`.
+2. Dereference is incomplete. Consider `int x = 5; int *ptr = &x;` and it is forbidden to use `*ptr`.
    However, it is valid to use `ptr[0]`, which behaves the same of `*ptr`.
-4. The support of varying number of function arguments is incomplete. No `<stdarg.h>` can be used.
+3. The support of varying number of function arguments is incomplete. No `<stdarg.h>` can be used.
```
jserv, line 170: Don't change the order. Minimize the required changes.

jserv: Later, we can create the corresponding issues via GitHub rather than the bullet points.

src/cfront.c
jserv: I agree. Let's stick to the reasonably effective changes.

src/cfront.c
```diff
@@ -1308,6 +1309,30 @@ int read_body_assignment(char *token, block_t *parent)
...
 }
 
+int read_global_assignment(char *token)
+{
+    var_t *var = find_global_var(token);
+    if (var) {
+        char buffer[10];
+        int isneg = 0;
+        /* global initialization must be constant */
+        /* TODO: support rvalue as expression e.g. int a = 3 + 2; */
```
jserv, line 1319: This issue should be resolved before this pull request is about to be merged.

eecheng87:
Latest commit support using expression to initialize global variable (also add test bench in `driver.sh`).

`read_global_assignment` using the similar concept with `read_expr` to process expression.
The reason I don't want merge them in one function is that they actually for totally different purpose.

`read_global_assignment` need to know the result before execution time and write into .data section before it.
`read_expr` don't need to get exact result of expression, it only need series instruction and get result during execution time.

src/cfront.c
```diff
@@ -1308,6 +1309,142 @@ int read_body_assignment(char *token, block_t *parent)
...
+    char buffer[10];
+    if (lex_accept(T_minus))
+        isneg = 1;
+    if (lex_peek(T_numeric, buffer))
+        res = read_numeric_constant(buffer);
+    else
+        error("Invalid value after assignment");
+    lex_expect(T_numeric);
+    if (isneg)
+        return -1 * res;
```
jserv, line 1325: Rewrite as `(-1) * res`.

src/cfront.c
```diff
@@ -1308,6 +1309,142 @@ int read_body_assignment(char *token, block_t *parent)
...
+    else
+        return res;
+}
+
+int get_expression_imm(opcode_t op, int op1, int op2)
```
jserv, line 1330: The verb `get` might not reflect what the function does. It can be `evaluate` or `eval` in short.

src/cfront.c
```diff
@@ -1308,6 +1309,142 @@ int read_body_assignment(char *token, block_t *parent)
...
+        break;
+    case OP_lshift:
+        res = op1 << op2;
+        break;
+    case OP_rshift:
+        res = op1 >> op2;
+        break;
+    default:
+        error("Using operation not support");
+        break;
```
jserv, line 1355: No `break` is required. Remove it.

README.md
```diff
@@ -165,8 +165,7 @@ int fib(int n)      fib:                        Reserve stack frame for function
 
 1. Any non-zero value is NOT treated as logical truth by all ops.
    That is, the expression `0 == strcmp(ptr, "hello")` is not equivalent to `!strcmp(ptr, "hello")`.
-2. Global variable initialization is not supported. 
-   Therefore, you can not initialize a global such as `int i = [expr]`.
+2. ELF lacks of .bss and .rodata section
```
jserv, line 168: Should be "The generated ELF file lacks..."

src/cfront.c
```diff
@@ -1308,6 +1309,140 @@ int read_body_assignment(char *token, block_t *parent)
...
+        res = op1 / op2;
+        break;
+    case OP_lshift:
+        res = op1 << op2;
+        break;
+    case OP_rshift:
+        res = op1 >> op2;
+        break;
+    default:
+        error("Using operation not support");
```
jserv, line 1353: Should be "The requested operation is not supported."

jserv (changes requested):
Improve the git commit message. You should address 1) how you made it in summary; 2) Known limitation; 3) New test item.

jserv (changes requested):
Change the message "Add correspond test bench in line 250 of driver.sh" to "Add global initialization item in test driver."

jserv (changes requested):
In git commit messages, always use quotes ("...") instead of backquote. Change "help us evaluate" to "evaluates."
Change "This commit doesn't" to "this patch ..."

jserv:

PR #16, opened 2020-10-16
Eliminate compilation warnings

sunshaoce:
to solve `warning: variable 'res' is used uninitialized whenever switch default is taken [-Wsometimes-uninitialized]`


jserv (changes requested):
Change the subject to something like "Eliminate compilation warnings."

jserv (changes requested):
You should change the subject of git commit message as well.

PR #19, opened 2020-10-25
Rework malloc and free function

eecheng87:
Rewrite "malloc" and add "free" function.
Import new data structure "block_meta_t"
to record extra information. 

Scheme for memory reuse in this patch is naive, just reuse whole block.
I'll implement better algorithm in future (e.g. split free block).

Corresponding example is tests/malloc_and_free.c

### DIFF
--- a/lib/c.c
+++ b/lib/c.c
@@ -370,12 +370,81 @@ int fputc(int c, FILE *stream)
  * FIXME: adopt lite_malloc from musl-libc
  *   https://git.musl-libc.org/cgit/musl/tree/src/malloc/lite_malloc.c
  */
-void *malloc(int size)
+
+typedef struct block_meta {
+    int size;
+    struct block_meta *next;
+    int free;
+} block_meta_t;
+
+block_meta_t *__malloc_global_base;
+block_meta_t *__malloc_global_last;
+
+block_meta_t *__malloc_request_space(int size)
 {
-    char *brk = __syscall(__syscall_brk, 0); /* read current break */
-    char *request = __syscall(__syscall_brk, brk + size);
+    char *brk;
+    block_meta_t *block;
+    brk = __syscall(__syscall_brk, 0);
+    block = brk;
+
+    char *request =
+        __syscall(__syscall_brk, block + size + sizeof(block_meta_t));
     if (request == -1)
         return NULL;
-    /* return previous location, now extended by size */
-    return brk;
+
+    if (__malloc_global_last != NULL)
+        __malloc_global_last->next = block;
+
+    block->size = size;
+    block->next = NULL;
+    block->free = 0;
+    return block;
+}
+
+void *malloc(int size)
+{
+    block_meta_t *block;
+    if (size == 0)
+        return NULL;
+    if (__malloc_global_base == NULL) {
+        block = __malloc_request_space(size);
+        if (block == NULL)
+            return NULL;
+        __malloc_global_base = block;
+    } else {
+        block_meta_t *current = __malloc_global_base;
+        __malloc_global_last = __malloc_global_base;
+        while (current != NULL) {
+            /* TODO: support break in while loop */
+            if (current->free == 1 && current->size >= size)
+                return current + 1;
+            __malloc_global_last = current;
+            current = current->next;
+        }
+        block = current;
+        if (block == NULL) {
+            block = __malloc_request_space(size);
+            if (block == NULL)
+                return NULL;
+        } else {
+            /* TODO: use the size requested instead of whole blocks */
+            block->free = 0;
+        }
+    }
+    return block + 1;
+}
+
+/* Quoting the C standard, 7.20.3.2/2 from ISO-IEC 9899:
+ * "If ptr is a null pointer, no action occurs."
+ */
+void free(void *ptr)
+{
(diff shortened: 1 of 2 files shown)


jserv:
Generally speaking, I would like to switch to mmap rather than brk system call as I described in #20. However, it would be still fine to improve brk based malloc/free as the transition.

tests/malloc_and_free.c
```diff
@@ -0,0 +1,18 @@
+int main(int argc, char *argv[])
```
jserv, line 1: Move the test program into `tests/driver.sh`.

jserv (changes requested):
The change was intended to rework malloc and free rather than "support."

tests/driver.sh
```diff
@@ -280,4 +280,18 @@ typedef enum { enum1 = 5, enum2 } enum_t;
 int main() { enum_t v = enum2; exit(v); }
 EOF
 
+# malloc and free
+try_ 1 << EOF
+int main()
+{
+    /* change test bench if different scheme apply */
+    int *a = malloc(sizeof(int) * 5);
```
jserv, line 288: Add `abort` if `a` equals to NULL.

lib/c.c
```diff
@@ -370,12 +370,89 @@ int fputc(int c, FILE *stream)
...
-void *malloc(int size)
+
+typedef struct block_meta {
+    int size;
+    struct block_meta *next;
+    int free;
+} block_meta_t;
+
+block_meta_t *global_base;
+block_meta_t *global_last;
```
jserv, line 381: Add pseudo namespace to avoid potential symbol conflicts. That is, rename the variables to something like `__malloc_global_base`.

tests/driver.sh
```diff
@@ -280,4 +280,20 @@ typedef enum { enum1 = 5, enum2 } enum_t;
...
+    free(a);
+    if (a == NULL)
+        abort();
+    int *b = malloc(sizeof(int) * 3);
+
+    /* "malloc" will reuse memory free by "free(a)" */
```
jserv, line 294: Change "free by" to "free'd by"

lib/c.c
```diff
@@ -370,12 +370,89 @@ int fputc(int c, FILE *stream)
...
+        } else {
+            /* TODO: use the size requested instead of whole blocks */
+            block->free = 0;
+        }
+    }
+    return block + 1;
+}
+
+void free(void *ptr)
+{
```
jserv, line 449: Add comment as following:
> Quoting the C standard, 7.20.3.2/2 from ISO-IEC 9899:
> "If ptr is a null pointer, no action occurs."

lib/c.c
```diff
@@ -370,12 +370,89 @@ int fputc(int c, FILE *stream)
...
+        block = request_space(size);
+        if (block == NULL)
+            return NULL;
+        __malloc_global_base = block;
+    } else {
+        __malloc_global_last = __malloc_global_base;
+        block = find_free_block(size);
+        if (block == NULL) {
+            block = request_space(size);
+            if (block == NULL) {
```
jserv, line 437: Change from `if (block == NULL) {` to `if (block == NULL)` for consistency.

lib/c.c
```diff
@@ -370,12 +370,89 @@ int fputc(int c, FILE *stream)
...
+typedef struct block_meta {
+    int size;
+    struct block_meta *next;
+    int free;
+} block_meta_t;
+
+block_meta_t *__malloc_global_base;
+block_meta_t *__malloc_global_last;
+
+block_meta_t *get_block_ptr(void *ptr)
```
jserv, line 383: Inline `get_block_ptr` into function `free`. That is, we should eliminate the necessary symbols exposed in libc.

lib/c.c
```diff
@@ -370,12 +370,89 @@ int fputc(int c, FILE *stream)
...
+        /* TODO: support break in while loop */
+        if (current->free == 1 && current->size >= size)
+            return current;
+        __malloc_global_last = current;
+        current = current->next;
+    }
+    return current;
+}
+
+block_meta_t *request_space(int size)
```
jserv, line 401: Function `request_space` would be called multiple times. Let's add pseudo namespace. That is, specify prefix `__malloc_` for this function to avoid potential symbol conflicts.

lib/c.c
```diff
@@ -370,12 +370,89 @@ int fputc(int c, FILE *stream)
...
+block_meta_t *__malloc_global_last;
+
+block_meta_t *get_block_ptr(void *ptr)
 {
-    char *brk = __syscall(__syscall_brk, 0); /* read current break */
-    char *request = __syscall(__syscall_brk, brk + size);
+    return ptr - sizeof(block_meta_t);
+}
+
+block_meta_t *find_free_block(int size)
```
jserv, line 388: Ditto. Inline `find_free_block` function into `main`.

eecheng87: Do you means inline `find_free_block` function into `malloc`?

jserv: Exactly. Please be aware of the fact that the content of `lib/c.c` will be inlined into C source files unless `-no-libc` is specified. Therefore, it is vital to avoid symbol confliction.

PR #23, opened 2020-10-28
Support break from general loop

### LINKED ISSUE #21: Doesn't support break in while loop
It is not valid now:

```
 while(1){
    break;
}
```

eecheng87:
Close #21


tests/driver.sh
```diff
@@ -296,4 +296,16 @@ int main()
 }
 EOF
 
+# break from while
+try_ 1 << EOF
+int main()
+{
+    while(1) {
+        break;
+        abort();
```
jserv, line 305: Add comment around `abort()`. For instance, "should not reach here."

tests/driver.sh
```diff
@@ -296,4 +296,17 @@ int main()
 }
 EOF
 
+# break from while
```
jserv, line 299: The existing section `loop` already contains `while(1) break; exit(10);`. You should enable it along with the similar cases.

tests/driver.sh
```diff
@@ -296,4 +296,17 @@ int main()
...
+try_ 1 << EOF
```
jserv, line 300: The following test program might be redundant if the test cases in `loop` section are enabled.

src/cfront.c
```diff
@@ -1514,24 +1514,35 @@ void read_body_statement(block_t *parent)
 
     if (lex_accept(T_while)) {
         ir_instr_t *false_jump;
-        ir_instr_t *start = add_instr(OP_label); /* start to return to */
-
+        ir_instr_t *start_jump =
```
jserv, line 1517: Can you confirm the case `for (;;) break;` as well?

eecheng87: should I send another PR? or also support `for (;;) break;` in this PR?

jserv: Let's consolidate the "break" support for both `while` and `for` loop in this pull request.

eecheng87: Now break is valid in `do_while`, `for` and `while` !

lazyparser: good job

PR #24, opened 2020-10-29
Support continue for general loop

eecheng87:
It is valid to use `continue` in `for`, `do_while` and `while` loop, also add new test bench in driver.sh


tests/driver.sh
```diff
@@ -122,8 +122,9 @@ items 60 "int acc; acc = 15; do { acc = acc * -2; } while (acc < 0); exit(acc);"
...
 # items 45 "int i; int j; i=0; j=0; while (i<10) { j=j+i; i=i+1; } exit(j);"
 items 1 "int x; x=0; do {x = x + 1; break;} while (1); exit(x);"
-# items 1 "int x; x=0; do {x = x + 1; continue;} while (0); exit(x);"
-# items 7 "int i; i=0; int j; for (j = 0; j < 10; j++) { if (j < 3) continue; i = i + 1; } exit(i);"
+items 2 "int x; x=0; do {x++; continue; abort();} while (x < 2); exit(x);"
+items 2 "int x; x=0; while(x < 2){x++; continue; abort();} exit(x);"
+items 7 "int i; i=0; int j; for (j = 0; j < 10; j++) { if (j < 3) continue; i = i + 1; } exit(i);"
 items 10 "while(0); exit(10);"
 items 10 "while(1) break; exit(10);"
 items 10 "for(;;) break; exit(10);"
```
jserv, line 130: Great!

You can add more test cases as following:
```c
items 30 "int i; int acc; i = 0; acc = 0; do { i = i + 1; if (i - 1 < 5) continue; acc = acc + i; if (i == 9) break; } while (i < 10); exit(acc);"
items 26 "int acc; acc = 0; int i; for (i = 0; i < 100; ++i) { if (i < 5) continue; if (i == 9) break; acc = acc + i; } exit(acc);"
```

eecheng87: The test case you mentioned is failing caused by no prefix addition support now.

Should I support prefix addition and subtraction in this PR?

jserv: Yes, please submit yet another pull request which does arithmetic operations.
You can simplify the above cases for passing.

jserv (changes requested):
You should rebase the latest master branch.

eecheng87:
> You should rebase the latest master branch.

Done, by 
> git pull --rebase upstream master
> git push --force origin pr

PR #25, opened 2020-10-31
Consolidate prefix operation

eecheng87:
This patch allows the prefix statement with l-value.
e.g. ++i;

Relative test benches in driver.sh are uncommented.


jserv (changes requested):
Fix the grammar errors such as uncomment"ed" and let"s."

jserv (changes requested):
Change the commit message to
> This patch allows the prefix statement with l-value.

PR #26, opened 2020-11-01
Support exclusive-or operator

eecheng87:
Support exclusive-or operator by adding instruction `eor`.
This patch also let's `get_operator_prio` finer to support more operator.


src/arm.c
```diff
@@ -22,6 +22,7 @@
 /* opcode */
 typedef enum {
     arm_and = 0,
+    arm_xor = 1,
```
jserv, line 25: The instruction is called `EOR` rather than `XOR`.
See https://developer.arm.com/documentation/dui0489/h/arm-and-thumb-instructions/and--orr--eor--bic--and-orn

src/cfront.c
```diff
@@ -846,19 +851,25 @@ void read_expr_operand(int param_no, block_t *parent)
...
-        return -2;
-
-    /* apply second last, low priority */
-    if ((op == OP_eq) || (op == OP_neq) || (op == OP_lt) || (op == OP_leq) ||
-        (op == OP_gt) || (op == OP_geq))
-        return -1;
-
-    /* apply first, high priority */
+    /* https://www.cs.uic.edu/~i109/Notes/COperatorPrecedenceTable.pdf */
+    if (op == OP_log_or)
```
jserv, line 855: You can rewrite with switch-case for readability.

src/arm.c
eecheng87: Actually, I use `EOR` unstruction. `XOR` is just for its naming convention. Should I change them to `eor` for naming?

jserv: You should use `EOR` for consistency with Arm reference manual.

src/arm.c
```diff
@@ -138,6 +139,11 @@ int __or_r(arm_cond_t cond, arm_reg rd, arm_reg rs, arm_reg rm)
     return __mov(cond, 0, arm_orr, 0, rs, rd, rm);
 }
 
+int __xor_r(arm_cond_t cond, arm_reg rd, arm_reg rs, arm_reg rm)
```
jserv, line 142: Should be `__eor_r`.

jserv (changes requested):
Change the commit message:
> This patch also extends "get_operator_prio" for more operators.

PR #27, opened 2020-11-05
Support modulo operator

eecheng87:


src/codegen.c
```diff
@@ -367,6 +368,14 @@ void code_generate()
             if (dump_ir == 1)
                 printf("    x%d /= x%d", dest_reg, OP_reg);
             break;
+        case OP_mod:
+            emit(__div(__AL, OP_reg + 1, OP_reg, dest_reg));
+            emit(__mul(__AL, OP_reg, OP_reg, OP_reg + 1));
+            emit(__sub_r(__AL, dest_reg, dest_reg, OP_reg));
+            /* TODO: support print '%' */
```
jserv, line 375: The comment was a bit misleading. Change it to 
>  TODO: support percent-sign character (%) in format string.

src/codegen.c
```diff
@@ -367,6 +368,14 @@ void code_generate()
             if (dump_ir == 1)
                 printf("    x%d /= x%d", dest_reg, OP_reg);
             break;
+        case OP_mod:
+            emit(__div(__AL, OP_reg + 1, OP_reg, dest_reg));
```
jserv, line 372: Both `div` and `mul` instruction are heavier. Instead, you can add shortcut for power-of-2. That is, convert the original modulo operation into `n & (d - 1)` if `d` is power-of-2. You should implement the shortcut in `src/cfront.c`.

eecheng87: I think only global expression can do this shortcut because `shecc` can't determine the value of operand in local expression before execution time. Can we ignore optimization for local expression in this patch?

BTW, relative [issue#28](https://github.com/jserv/shecc/issues/28) and detailed information was sent.

jserv: Sure, let's merge this change first. Do append the comments with `TODO: optimize ...`

PR #29, opened 2020-11-07
Support bitwise not operator

eecheng87:


jserv:
Don't check in the generated file `bench` again. Send another pull request.

PR #31, opened 2020-11-08
Consolidate logical not operator

eecheng87:
Fix known issue 1 and un-comment relative test bench in `driver.sh`

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,10 +164,6 @@ int fib(int n)      fib:                        Reserve stack frame for function
 
 ## Known Issues
 
-1. The compiler uses different logic to evaluate the operand of `if` or `!` than a general integer
-   expression. Therefore, explicit failback to "evaluate value and compare against zero" is required
-   if the operand is not a "truthiness-generating" expression.
-   That is, the expression `0 == strcmp(ptr, "hello")` is not equivalent to `!strcmp(ptr, "hello")`.
 2. The generated ELF lacks of .bss and .rodata section
 3. The unary `*` operator is not supported, which makes it necessary to use `[0]` syntax.
    Consider `int x = 5; int *ptr = &x;` and it is forbidden to use `*ptr`. However, it is valid to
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -164,21 +164,17 @@ int fib(int n)      fib:                        Reserve stack frame for function
...
-   expression. Therefore, explicit failback to "evaluate value and compare against zero" is required
-   if the operand is not a "truthiness-generating" expression.
-   That is, the expression `0 == strcmp(ptr, "hello")` is not equivalent to `!strcmp(ptr, "hello")`.
-2. The generated ELF lacks of .bss and .rodata section
-3. The unary `*` operator is not supported, which makes it necessary to use `[0]` syntax.
+1. The generated ELF lacks of .bss and .rodata section
+2. The unary `*` operator is not supported, which makes it necessary to use `[0]` syntax.
    Consider `int x = 5; int *ptr = &x;` and it is forbidden to use `*ptr`. However, it is valid to
    use `ptr[0]`, which behaves the same of `*ptr`.
-4. The support of varying number of function arguments is incomplete. No `<stdarg.h>` can be used.
```
jserv, line 175: You don't have to change the numbers. Markdown parser can still retrieve from non-continuous numbers. You can simply remove the first item.

jserv (changes requested):
It is "logical not" operator instead of "logic not."

src/codegen.c
```diff
@@ -468,10 +468,11 @@ void code_generate()
             if (dump_ir == 1)
                 printf("    x%d >>= x%d", dest_reg, OP_reg);
             break;
-        case OP_not:
+        case OP_log_not:
             /* 1 if zero, 0 if nonzero */
-            /* only works for 1/0 */
-            emit(__rsb_i(__AL, dest_reg, 1, dest_reg));
+            emit(__teq(dest_reg));
```
jserv, line 473: Is it feasible to avoid the setting for both `0` and `1` every time we counter logical not?

eecheng87: Yes it is feasible, but code length must be 12. Code is update~

PR #32, opened 2020-11-08
Clean up the comparison with zero

jserv:
This patch replaced all occurrences of "0 == strcmp(a, b)" with
"!strcmp(a, b)". Also, all expressions containing "!= NULL" were changed
into the simplified form.

### DIFF
--- a/lib/c.c
+++ b/lib/c.c
@@ -315,9 +315,9 @@ void abort()
 
 FILE *fopen(char *filename, char *mode)
 {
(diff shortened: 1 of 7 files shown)


jserv:
TODO: the expression `!= 0` can be eliminated as well. Check the statement `blk->locals[i].init_val != 0` in file src/codegen.c.

eecheng87 (approved):

PR #34, opened 2020-11-14
Reserve return value from main function

eecheng87:


jserv (changes requested):
Rewrite all occurrence of exit(?) in the end of test items with the ordinary `return` statement.

tests/driver.sh
```diff
@@ -140,7 +140,7 @@ int sum(int m, int n) {
     int i;
     for (i = m; i <= n; i = i + 1)
         acc = acc + i;
-    exit(acc);
+    return(acc);
```
jserv, line 143: Simply use `return acc` without extra parentheses.

PR #35, opened 2020-11-25
Support ternary operator

eecheng87:
This patch support ternary operator in:
1. global variable initialization
2. local variable initialization & assignment
3. expression of return
4. parameter list of function call

This patch also increases limitations in `defs.h` because size of shecc grows a lot!    (See #36 )

### DIFF
--- a/src/cfront.c
+++ b/src/cfront.c
@@ -73,6 +73,7 @@ typedef enum {
     T_assign,        /* = */
     T_increment,     /* ++ */
     T_decrement,     /* -- */
+    T_question,      /* ? */
     T_colon,         /* : */
     T_semicolon,     /* ; */
     T_eof,           /* end-of-file (EOF) */
@@ -416,6 +417,10 @@ token_t get_next_token()
         read_char(1);
         return T_semicolon;
     }
+    if (next_char == '?') {
+        read_char(1);
+        return T_question;
+    }
     if (next_char == ':') {
         read_char(1);
         return T_colon;
@@ -697,12 +702,14 @@ void read_char_param(int param_no)
     ii->int_param1 = token[0];
 }
 
+void read_ternary_operation(int dest, block_t *parent);
 void read_func_parameters(block_t *parent)
 {
     int param_num = 0;
     lex_expect(T_open_bracket);
     while (!lex_accept(T_close_bracket)) {
         read_expr(param_num++, parent);
+        read_ternary_operation(param_num - 1, parent);
         lex_accept(T_comma);
     }
 }
@@ -871,6 +878,8 @@ int get_operator_prio(opcode_t op)
 {
     /* https://www.cs.uic.edu/~i109/Notes/COperatorPrecedenceTable.pdf */
     switch (op) {
+    case OP_ternary:
+        return 3;
     case OP_log_or:
         return 4;
     case OP_log_and:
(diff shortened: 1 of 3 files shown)


jserv:
> This patch also increases limitations in `defs.h` because size of `shecc` grows a lot!

We should create another issue to address the memory consumption.

tests/driver.sh
```diff
@@ -259,8 +259,14 @@ int main()
 EOF
 
 # conditional operator
-# expr 10 "1 ? 10 : 5"
```
jserv, line 262: Why didn't you re-use the existing test items?

eecheng87: Because it is not complicated enough and it only cover "conditional operator in return statement". Current test bench covers three kind of usages.

eecheng87:
> > This patch also increases limitations in `defs.h` because size of `shecc` grows a lot!
> 
> We should create another issue to address the memory consumption.

Do I need to create another PR to link relative issue or just link to that in this PR?

tests/driver.sh
jserv: For profiling purpose, it would be better to have minimized test items available. Thus, it can reduce uncertainty.

jserv:
> > We should create another issue to address the memory consumption.
> Do I need to create another PR to link relative issue or just link to that in this PR?

Simply describe in new issue(s).

tests/driver.sh
eecheng87: Test bench was reverted!

eecheng87:
> > > We should create another issue to address the memory consumption.
> > > Do I need to create another PR to link relative issue or just link to that in this PR?
> 
> Simply describe in new issue(s).

Issue was linked!

src/cfront.c
```diff
@@ -1425,28 +1497,39 @@ int read_global_assignment(char *token)
         int val_stack[10];
         int op_stack_index = 0, val_stack_index = 0;
         int operand1, operand2;
+        int cond = -1;
```
jserv, line 1500: Can you reduce the scope of `cond`? That is, declare it in the later block.

jserv (changes requested):
The subject should be adjusted. It was meant to ternary operator.
See https://en.wikipedia.org/wiki/%3F:

src/cfront.c
```diff
@@ -1269,6 +1280,35 @@ void read_lvalue(lvalue_t *lvalue,
     }
 }
 
+void read_ternary_operation(int dest, block_t *parent)
+{
+    if (lex_accept(T_question)) {
```
jserv, line 1285: You can rewrite with false-condition-first manner. That is,
```c
if (!lex_accept(T_question)) return;

/* the reset code */
```

Then, the program can be shortened.

src/cfront.c
```diff
@@ -1269,6 +1280,35 @@ void read_lvalue(lvalue_t *lvalue,
...
+        /* ternary-operator */
+        ir_instr_t *false_jump;
+        false_jump = add_instr(OP_jz);
+        false_jump->param_no = dest;
+
+        /* true branch */
+        read_expr(dest, parent);
+        if (lex_accept(T_colon)) {
```
jserv, line 1293: Ditto. Follow false-condition-first manner.

tests/driver.sh
jserv: You can append the new ternary operation test case as well.

PR #39, opened 2020-12-30
Fix missing terminating " character error in libc generated by inliner

HDB-PROGRAMMING:
When I was trying to build shecc, my compiler just threw me an `missing terminating " character` in the libc build stage, so I update the inliner to write a `\` character at the end of every line. Now it just throws me an `Segmentation fault (core dumped)` runtime error in the compiler executable (I'll try to fix it too).

### DIFF
--- a/tools/inliner.c
+++ b/tools/inliner.c
@@ -28,17 +28,27 @@ void write_str(char *str)
 
 void write_line(char *src)
 {
-    int i;
-
+    int i,j;
     write_str("  __c(\"");
+
     for (i = 0; src[i]; i++) {
         if (src[i] == '\"') {
             write_char('\\');
             write_char('\"');
         } else if (src[i] != '\n') {
             write_char(src[i]);
+        } else if (src[i] == '\n') {
+            write_char('\\');
+            write_char('\n');
         }
     }
+
+    for (j = 0; SOURCE[j]; j++) {
+        if (SOURCE[j] == '\r') {
+            SOURCE[j] = ' ';
+        }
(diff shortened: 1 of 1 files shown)


jserv:
@HDB-PROGRAMMING, can you provide more information about your build environment? e.g. operating system and compiler you are using.

HDB-PROGRAMMING:
@jserv Yes, I'm using Windows WSL 2 with GCC compiler

jserv:
> @jserv Yes, I'm using Windows WSL 2 with GCC compiler. Pd: your project is pretty cool

That is why you encountered the newline character issue.
* Unix and Unix-like systems: "\n"
* Microsoft Windows, DOS: "\r\n"
* macOS: "\r"

See: https://en.wikipedia.org/wiki/Newline#Representation

HDB-PROGRAMMING:
@jserv Which OS do I need to use to build shecc?

eecheng87:
> @jserv Which OS do I need to use to build shecc?

It works fine under Ubuntu 20.04.1 LTS.

HDB-PROGRAMMING:
@eecheng87 Thanks, at least I think my PR could be useful in some OSs

tools/inliner.c
```diff
@@ -28,17 +28,27 @@ void write_str(char *str)
 
 void write_line(char *src)
 {
-    int i;
-
+    int i,j;
```
jserv, line 31: Use `clang-format -i` to ensure the consistent style.

Issue #10, opened 2020-10-12
RISC-V code generation


jserv:
Per the request of @lazyparser, `shecc` would support RISC-V code generation and still self-host.

lazyparser:
Thanks ;-)

is there any guidelines for new backends contribution?

I'd like to contribute

MikeDB1:
I'd also comment that ARM 64 bit support would be useful.  The Pi4 is much more efficient on 64bit and some processors are already 64 bit only

eecheng87:
Hi, I try to contribute this feature recently. Before started, I want to make sure that my modification is under your expectation and approval. Following is high level description about what I'll do:

* Add parameter (argc/argv) to decide target (RISC-V or ARM)
* Modify code in `codegen.c` to support emitting ARM or RISC-V instruction
* Add new file `riscv.c` to mapping instruction emit by `codegen.c` into RISC-V bytecode

Also, I've some question about how to verify result compiled in RISC-V bytecode? Is it as same as how to verify ARM-based result? Are there any RISC-V simulator you like me to use as we run ARM bytecode on `qemu`.

Any suggestion is appreciated!

lazyparser:
> Hi, I try to contribute this feature recently. Before started, I want to make sure that my modification is under your expectation and approval. Following is high level description about what I'll do:
> 
>     * Add parameter (argc/argv) to decide target (RISC-V or ARM)
> 
>     * Modify code in `codegen.c` to support emitting ARM or RISC-V instruction
> 
>     * Add new file `riscv.c` to mapping instruction emit by `codegen.c` into RISC-V bytecode
> 
> 
> Also, I've some question about how to verify result compiled in RISC-V bytecode? Is it as same as how to verify ARM-based result? Are there any RISC-V simulator you like me to use as we run ARM bytecode on `qemu`.
> 
> Any suggestion is appreciated!

Fantastic!

Spike and QEMU are both the emulator we daily use.

https://github.com/riscv/riscv-isa-sim is the repo of spike.

The easiest way to deploly QEMU on linux/ubuntu can ref:
https://github.com/isrc-cas/PLCT-Toolbox/blob/master/deploy_riscv64fedora_qemu.sh

for a user-level binary, the user mode QEMU is very handy. the usage of qemu-riscv64/qemu-riscv32 is much like the QEMU-arm shecc used. 

For the RISC-V, currently RV64G is more mature than the RV32G platform. But for an educational compiler, RV32I might be good & precise enough.

[...]

PR #40, opened 2021-01-14
Implement RISC-V backend

eecheng87:
This patch rework risc-v backend by fixing several bugs and adding
several new features

Fix:
- Inproper value return from main function
- Wrong length of B-type instruction
- Fail to import built-in libc
- Can't open file
- Can't pass stage2 ,bootstrap and driver.sh

Add:
- xor
- bit not
- mod
- div

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ a.out
 *.elf
 *.log
 *.lst
+config
+src/codegen.c
--- a/Makefile
+++ b/Makefile
@@ -26,12 +26,15 @@ all: bootstrap
 
 config:
 ifeq (riscv,$(ARCH))
-	@echo $(RISCV_EXEC) > $(OUT)/target
+	@$(VECHO) "$(RISCV_EXEC)" > $(OUT)/target
+	@$(VECHO) "#define TARGET_RISCV 1" > $@
+	@ln -s $(PWD)/$(SRCDIR)/riscv-codegen.c $(SRCDIR)/codegen.c
 else
-	@echo $(ARM_EXEC) > $(OUT)/target
+	@$(VECHO) "$(ARM_EXEC)" > $(OUT)/target
+	@$(VECHO) "#define TARGET_ARM 1" > $@
+	@ln -s $(PWD)/$(SRCDIR)/arm-codegen.c $(SRCDIR)/codegen.c
 endif
-	@echo -n Target machine code switch to
-	@echo `cat $(OUT)/target` | sed 's/.*qemu-\([^ ]*\).*/ \1/'
+	@$(VECHO) "Target machine code switch to %s\n" "$$(cat out/target | sed 's/.*qemu-\([^ ]*\).*/\1/')"
 
 $(OUT)/tests/%.elf: tests/%.c $(OUT)/$(STAGE0)
 	$(VECHO) "  SHECC\t$@\n"
@@ -80,6 +83,6 @@ clean:
 	-$(RM) $(OBJS) $(deps)
 	-$(RM) $(TESTBINS) $(OUT)/tests/*.log $(OUT)/tests/*.lst
 	-$(RM) $(OUT)/shecc*.log
-	-$(RM) $(OUT)/inliner $(OUT)/libc.inc $(OUT)/target
+	-$(RM) $(OUT)/inliner $(OUT)/libc.inc $(OUT)/target config $(SRCDIR)/codegen.c
 
 -include $(deps)
--- a/lib/c.c
+++ b/lib/c.c
@@ -11,6 +11,16 @@
 #define __syscall_brk 45
 #endif
 
+#ifdef __riscv
+#define __syscall_exit 93
+#define __syscall_read 63
+#define __syscall_write 64
+#define __syscall_close 57
+#define __syscall_open 1024
+#define __syscall_openat 56
+#define __syscall_brk 214
+#endif
+
 typedef int FILE;
 
 void abort();
@@ -316,9 +326,21 @@ void abort()
 FILE *fopen(char *filename, char *mode)
 {
     if (!strcmp(mode, "wb"))
+#ifdef __arm__
         return __syscall(__syscall_open, filename, 65, 0x1fd);
+#endif
+/* FIXME: mode not work currently in RISC-V */
+#ifdef __riscv
+    return __syscall(__syscall_openat, -100, filename, 65, 0x1fd);
+#endif
     if (!strcmp(mode, "rb"))
+#ifdef __arm__
         return __syscall(__syscall_open, filename, 0, 0);
+#endif
+#ifdef __riscv
+    return __syscall(__syscall_openat, -100, filename, 0, 0);
+#endif
+
     abort();
 }
 
--- a/src/cfront.c
+++ b/src/cfront.c
@@ -2145,8 +2145,15 @@ void parse_internal()
     add_block(NULL, NULL);    /* global block */
     elf_add_symbol("", 0, 0); /* undef symbol */
 
-    /* architecture defines */
+/* architecture defines */
+/* FIXME: use #ifdef ... #else ... #endif */
+#ifdef TARGET_ARM
     add_alias("__arm__", "1"); /* defined by GNU C and RealView */
+#endif
+#ifdef TARGET_RISCV
+    /* Older versions of the GCC toolchain defined __riscv__ */
+    add_alias("__riscv", "1");
+#endif
 
     /* binary entry point: read params, call main, exit */
     ii = add_instr(OP_label);
--- a/src/elf.c
+++ b/src/elf.c
@@ -82,7 +82,13 @@ void elf_generate_header()
     elf_write_header_int(0);          /* EI_PAD: unused */
     elf_write_header_byte(2);         /* ET_EXEC */
     elf_write_header_byte(0);
+/* FIXME: use #ifdef ... #else ... #endif */
+#ifdef TARGET_ARM
     elf_write_header_byte(0x28); /* ARM (up to ARMv7/Aarch32) */
+#endif
+#ifdef TARGET_RISCV
+    elf_write_header_byte(0xf3); /* RISC-V */
+#endif
     elf_write_header_byte(0);
     elf_write_header_int(1);                          /* ELF version */
     elf_write_header_int(ELF_START + elf_header_len); /* entry point */
@@ -90,9 +96,15 @@ void elf_generate_header()
     elf_write_header_int(elf_header_len + elf_code_idx + elf_data_idx + 39 +
                          elf_symtab_index +
                          elf_strtab_index); /* section header offset */
-    /* flags */
+/* flags */
+/* FIXME: use #ifdef ... #else ... #endif */
+#ifdef TARGET_ARM
     elf_write_header_int(0x5000200); /* ARM */
-    elf_write_header_byte(0x34);     /* header size */
+#endif
+#ifdef TARGET_RISCV
+    elf_write_header_int(0);
+#endif
+    elf_write_header_byte(0x34); /* header size */
     elf_write_header_byte(0);
     elf_write_header_byte(0x20); /* program header size */
     elf_write_header_byte(0);
--- a/src/main.c
+++ b/src/main.c
@@ -4,6 +4,9 @@
 #include <stdlib.h>
 #include <string.h>
 
+/* Define target machine */
+#include "../config"
+
 /* The inclusion must follow the fixed order, otherwise it fails to build. */
 #include "defs.h"
 
@@ -17,7 +20,7 @@
 #include "cfront.c"
 
 /* Machine code generation. support ARMv7-A and RISC-V32I */
-#include "arm-codegen.c"
+#include "codegen.c"
 
 /* inlined libc */
 #include "../out/libc.inc"
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,8 @@ script:
   - make config ARCH=arm
   - make VERBOSE=1      
   - make check
+  - make clean
+  - make config ARCH=riscv
+  - make VERBOSE=1
+  - make check
   - sh .ci/check-format.sh
--- a/src/riscv-codegen.c
+++ b/src/riscv-codegen.c
@@ -1,0 +1,596 @@
+/* Translate IR to target machine code */
+
+#include "riscv.c"
+
+/* Compute stack space needed for function's parameters */
+void size_func(func_t *fn)
+{
+    int s = 0, i;
+
+    /* parameters are turned into local variables */
+    for (i = 0; i < fn->num_params; i++) {
+        s += size_var(&fn->param_defs[i]);
+        fn->param_defs[i].offset = s; /* stack offset */
+    }
+
+    /* align to 16 bytes */
+    if ((s & 15) > 0)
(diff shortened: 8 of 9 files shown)


eecheng87:
It works fine in local. Travis CI can't use `qemu-riscv32` even after installing `qemu-user`.

lib/c.c
```diff
@@ -325,9 +326,21 @@ void abort()
...
+/* FIXME: mode not work currently in riscv */
+#ifdef __riscv
+    return __syscall(__syscall_openat, -100, filename, 65, 0x1fd);
+#endif
     if (!strcmp(mode, "rb"))
+#ifdef __arm__
         return __syscall(__syscall_open, filename, 0, 0);
+#endif
+#ifdef __riscv
+    return __syscall(__syscall_openat, -100, filename, 0, 0);
```
jserv, line 341: I am not sure if this change is necessary. Can you explain?

eecheng87: rv32 doesn't support system call `open` (at least 1024 isn't the system call number for `open`). Also, in this [ref](https://fedora.juszkiewicz.com.pl/syscalls.html) show that rv32 doesn't support `open`.

jserv:
Let's prepare the following pull request(s) prior to the RISC-V backend support:
1. Decouple Arm backend and make it switchable. It is not necessary to provide run-time switch, but we do need to introduce another layer for both Arm and RISC-V backend.
2. Rename the existing source files. That is, `src/codegen.c` becomes `src/arm-codegen.c`, and there should be a symbolic `src/codegen.c` to the corresponding backend configuration.
3. Make Travis-CI ready for testing. We might migrate to [GitHub Actions](https://github.com/features/actions) later.

lib/c.c
jserv: Append the above information within the comments.

jserv: I would be afraid of the conditional compilation, which looks a bit ugly without `#else` and/or `#elif`.

jserv:
> It works fine in local. The reason of failing pass is that Travis CI can't use `qemu-riscv32` even after installing `qemu-user`.

You should make a list of QEMU versions you have been successfully tested. I suffered from some broken user-mode emulation for earlier QEMU though.

lazyparser:
> > It works fine in local. The reason of failing pass is that Travis CI can't use `qemu-riscv32` even after installing `qemu-user`.
> 
> You should make a list of QEMU versions you have been successfully tested. I suffered from some broken user-mode emulation for earlier QEMU though.

4.2.0 and 5.0.0 would be good for testing.

btw, if you met any issues with QEMU, my colleagues working on QEMU/Spike would happy to help. feel  free to open issues.

eecheng87:
> > > It works fine in local. The reason of failing pass is that Travis CI can't use `qemu-riscv32` even after installing `qemu-user`.
> > 
> > 
> > You should make a list of QEMU versions you have been successfully tested. I suffered from some broken user-mode emulation for earlier QEMU though.
> 
> 4.2.0 and 5.0.0 would be good for testing.
> 
> btw, if you met any issues with QEMU, my colleagues working on QEMU/Spike would happy to help. feel free to open issues.

Thanks for reply, I think the version of `qemu-user` is new enough in local (above 4.2.0).
It seems that the default version of `qemu-user` installed in TravisCI is too old to use `qemu-riscv32` and 
maybe we can't installed `qemu-user` with specific version such as:
```
    packages:
      - qemu-user=4.2.0
``` 
Thus, I've no idea how to execute `qemu-riscv32` in TravisCI.

eecheng87:
> Let's prepare the following pull request(s) prior to the RISC-V backend support:
> 
> 1. Decouple Arm backend and make it switchable. It is not necessary to provide run-time switch, but we do need to introduce another layer for both Arm and RISC-V backend.

Can you describe more about "another layer for both Arm and RISC-V backend" and how do we decide which backend to use? Using argv?

Thanks.

jserv:
> Can you describe more about "another layer for both Arm and RISC-V backend" and how do we decide which backend to use? Using argv?

Instead of executing `$(ARM_EXEC)` in Makefile, we can use indirect invocation such as `$(TARGET_EXEC)` where `TARGET` refers to the pre-configured target. In both build system and source tree, we should avoid the lousy composition of Arm and RISC-V mixture.

jserv:
> It seems that the default version of `qemu-user` installed in TravisCI is too old to use `qemu-riscv32` and
> maybe we can't installed `qemu-user` with specific version such as:

Travis-CI supports Ubuntu 20.04 (focal): https://docs.travis-ci.com/user/reference/overview/
Thus, you can send another pull request to upgrade the base Ubuntu Linux system and then install the preferable QEMU user emulation.

eecheng87:
> > It seems that the default version of `qemu-user` installed in TravisCI is too old to use `qemu-riscv32` and
> > maybe we can't installed `qemu-user` with specific version such as:
> 
> Travis-CI supports Ubuntu 20.04 (focal): https://docs.travis-ci.com/user/reference/overview/
> Thus, you can send another pull request to upgrade the base Ubuntu Linux system and then install the preferable QEMU user emulation.

Thanks for your reply, `qemu-riscv32` works good under Ubuntu 20.04 in Travis-CI !

lib/c.c
eecheng87: Got it, I will append above information. Also, I'll send a issue about lack of feature `#else` and `#elif`. It seems fine to use only `#ifdef` now. Feature about `#else` and `#elif` will be added in future PR.

jserv:
Please rework these git commits and rebase the latest `master` branch. You can drop my previous commits and use "Co-authored commits" feature instead: https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

eecheng87:
How about default codegen header in `main.c`?
```c
/* Machine code generation */
#include "riscv-codegen.c"
```
Let users to change themselves?

jserv:
> How about default codegen header in `main.c`?
> ```c
> /* Machine code generation */
> #include "riscv-codegen.c"
> ```
> Let users to change themselves?

No, after executing `make config`, the build system should create the symbolic link from {arm,riscv}-codegen.c to `codegen.c`. And then, file `main.c` should include only `codegen.c` without specifying the target.

jserv:
> ```c
>     if (0) {
>         elf_write_header_byte(0x28); /* ARM (up to ARMv7/Aarch32) */
>     } else {
>         elf_write_header_byte(0xf3); /* RISC-V */
>     }
> ```

That is the reason why we need the limited `#if` and `#else` support.

jserv:
> > That is the reason why we need the limited `#if` and `#else` support.
> Did this mean that we need to add this feature before this PR?

It would be better but not necessary.

eecheng87:
> It would be better but not necessary.

I'll add this in future, just only use `#ifdef` in this PR.

src/cfront.c
```diff
@@ -1,4 +1,5 @@
 /* C language front-end */
+/*#include "def.h"*/
```
jserv, line 2: Don't do this.

src/cfront.c
```diff
@@ -2145,8 +2146,14 @@ void parse_internal()
     add_block(NULL, NULL);    /* global block */
     elf_add_symbol("", 0, 0); /* undef symbol */
 
-    /* architecture defines */
+/* architecture defines */
```
jserv, line 2149: Add "FIXME: use #ifdef ... #else ... #endif"

src/elf.c
```diff
@@ -1,4 +1,5 @@
 /* ELF file manipulation */
+/*#include "def.h"*/
```
jserv, line 2: Don't do this.

eecheng87:
> You can drop my previous commits and use "Co-authored commits" feature instead

It seems that delete previous commit will also affect later commits. I think it might be squash my commits to your previous commit and I become co-authored?

src/cfront.c
eecheng87: Oh, I miss it

Makefile
```diff
@@ -27,8 +27,12 @@ all: bootstrap
 config:
 ifeq (riscv,$(ARCH))
 	@echo $(RISCV_EXEC) > $(OUT)/target
+	@echo \#define riscv 1 > $(SRCDIR)/def.h
```
jserv, line 30: Do not use `arm` and `riscv` for predefined macro names. Instead, use `TARGET_ARM` and `TARGET_RISCV`.

Makefile
```diff
@@ -27,8 +27,12 @@ all: bootstrap
 config:
 ifeq (riscv,$(ARCH))
 	@echo $(RISCV_EXEC) > $(OUT)/target
+	@echo \#define TARGET_RISCV 1 > $(SRCDIR)/def.h
```
jserv, line 30: `def.h` is quite confusing since we already have file `defs.h`.

jserv: We can generate the file `config.h` in the top-level directory rather than `$(SRCDIR)`.

jserv: It would be similar to the way what autotools do.

eecheng87: how about `target.h`?

jserv: `target.h` is confusing because it usually refers to target-specific configurations and entries. See LLVM for example.

Makefile
```diff
@@ -27,8 +27,12 @@ all: bootstrap
...
+	@echo \#include \"riscv-codegen.c\" > $(SRCDIR)/codegen.c
 else
 	@echo $(ARM_EXEC) > $(OUT)/target
+	@echo \#define TARGET_ARM 1 > $(SRCDIR)/def.h
+	@echo \#include \"arm-codegen.c\" > $(SRCDIR)/codegen.c
```
jserv, line 35: Don't use `echo`. Use symbolic link. i.e. `ln -s arm-codegen.c codegen.c`.

jserv:
> It seems that delete previous commit will also affect later commits. I think it might be squash my commits to your previous commit and I become co-authored?

Squashing is pretty well. Do it.

src/main.c
```diff
@@ -7,6 +7,9 @@
 /* The inclusion must follow the fixed order, otherwise it fails to build. */
 #include "defs.h"
 
+/* Define target machine */
+#include "def.h"
```
jserv, line 11: If the generated configurations are located in file `config.h`, then you should move it to the top of user headers.

jserv: i.e.
```c
#include "config.h"

#include "globals.c"
#include "cfront.c"
```

Makefile
eecheng87: OK, let's use `config.h`

eecheng87:
All commit was squashed into one, also requests from you were changed!

jserv (changes requested):
Add the generated "config.h" in gitignore. When executing "make distclean", file "config.h" should be removed.

Makefile
```diff
@@ -27,8 +27,12 @@ all: bootstrap
 config:
 ifeq (riscv,$(ARCH))
 	@echo $(RISCV_EXEC) > $(OUT)/target
+	@echo \#define TARGET_RISCV 1 > config.h
```
jserv, line 30: Use `$(VECHO)` rather than plain `echo`.

Makefile
```diff
@@ -27,8 +27,12 @@ all: bootstrap
...
+	@ln -s riscv-codegen.c $(SRCDIR)/codegen.c
```
jserv, line 31: Should be `$(SRCDIR)/riscv-codegen.c`.

Makefile
```diff
@@ -27,8 +27,12 @@ all: bootstrap
...
 else
 	@echo $(ARM_EXEC) > $(OUT)/target
+	@echo \#define TARGET_ARM 1 > config.h
+	@ln -s arm-codegen.c $(SRCDIR)/codegen.c
```
jserv, line 35: Ditto.

jserv: You might need stamp for `config` target.

Makefile
eecheng87: `$(SRCDIR)/riscv-codegen.c` will fail to include `codegen.c`, so I change to `riscv-codegen.c` and it works good.

lib/c.c
```diff
@@ -316,9 +326,21 @@ void abort()
 FILE *fopen(char *filename, char *mode)
 {
     if (!strcmp(mode, "wb"))
+#ifdef __arm__
         return __syscall(__syscall_open, filename, 65, 0x1fd);
+#endif
+/* FIXME: mode not work currently in riscv */
```
jserv, line 332: Mention "RISC-V" instead of "riscv" in comments.

Makefile
eecheng87: Do you means create a variable for "config" ?

jserv (changes requested):
Append "Close #10" in the end of git commit message.

Makefile
jserv: No, we should avoid running "make config" twice. Only the first time makes sense. For the second and more times of invoking "make config", there is no effect.

Makefile
jserv: You should figure out why it failed.

Makefile
jserv: Alternatively, you can replace `echo \#define TARGET_RISCV 1 > config.h` with `echo \#define TARGET_RISCV 1 > config` and include `config` in C source. Then, the build system would track the dependency of `config`.

eecheng87: Got it.

Makefile
eecheng87: How about this:
```shell
ln -s $(PWD)/$(SRCDIR)/arm-codegen.c $(SRCDIR)/codegen.c
```
It also works good.

.gitignore
```diff
@@ -4,3 +4,5 @@ a.out
 *.elf
 *.log
 *.lst
+config
+codegen.c
```
jserv, line 8: Should be `src/codegen.c`

eecheng87: Got it.

Makefile
```diff
@@ -26,12 +26,16 @@ all: bootstrap
...
+	@ln -s $(PWD)/$(SRCDIR)/riscv-codegen.c $(SRCDIR)/codegen.c
 else
-	@echo $(ARM_EXEC) > $(OUT)/target
+	@$(VECHO) "$(ARM_EXEC)" > $(OUT)/target
+	@$(VECHO) "#define TARGET_ARM 1" > config
+	@ln -s $(PWD)/$(SRCDIR)/arm-codegen.c $(SRCDIR)/codegen.c
 endif
-	@echo -n Target machine code switch to
-	@echo `cat $(OUT)/target` | sed 's/.*qemu-\([^ ]*\).*/ \1/'
+	@$(VECHO) "Target machine code switch to"
```
jserv, line 37: Can you merge the two lines?

Makefile
```diff
@@ -26,12 +26,16 @@ all: bootstrap
 
 config:
 ifeq (riscv,$(ARCH))
-	@echo $(RISCV_EXEC) > $(OUT)/target
+	@$(VECHO) "$(RISCV_EXEC)" > $(OUT)/target
+	@$(VECHO) "#define TARGET_RISCV 1" > config
```
jserv, line 30: Replace `> config` with `> $@`

Makefile
eecheng87: Sure, I can.

Makefile
eecheng87: Got it.

PR #41, opened 2021-01-14
Upgrade Travis-CI env to Ubuntu 20.04

eecheng87:
`qemu-user` under Ubuntu 18.04 in Travis-CI doesn't support `qemu-riscv32`, as
a result, we upgrade it to Ubuntu 20.04 (focal)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
 
 dist:
-  - bionic
+  - focal
 
 compiler:
   - gcc


.travis.yml
```diff
@@ -4,23 +4,25 @@ os:
...
   - gcc
 
 addons:
   apt:
     sources:
-      - sourceline: "deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ bionic-updates main universe"
+      - sourceline: 'ppa:chris-lea/redis-server'
     packages:
-      - qemu-user
+    - redis-tools
```
jserv, line 17: Why are redis tools and server included?

eecheng87: Removed

.travis.yml
```diff
@@ -4,15 +4,15 @@ os:
...
+  - focal
 
 compiler:
   - gcc
 
 addons:
   apt:
     sources:
-      - sourceline: "deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ bionic-updates main universe"
+      - sourceline: 'ppa:chris-lea/redis-server'
```
jserv, line 15: Is this change necessary?

eecheng87: No, removed it.

Issue #47, opened 2021-02-04
Refactor architecture specific configurations


jserv:
Instead of #43, I would propose an elegant way to specify the architecture and its configurations.  The WIP changes are shown as following:
```diff
diff --git a/Makefile b/Makefile
index 9d3e9ca..8a31660 100644
--- a/Makefile
+++ b/Makefile
@@ -20,21 +20,22 @@ OBJS := $(SRCS:%.c=$(OUT)/%.o)
 deps := $(OBJS:%.o=%.o.d)
 TESTS := $(wildcard tests/*.c)
 TESTBINS := $(TESTS:%.c=$(OUT)/%.elf)
-TARGET_EXEC := `cat $(OUT)/target`
 
 all: config bootstrap
 
-config:
-ifeq (riscv,$(ARCH))
-	@$(VECHO) "$(RISCV_EXEC)" > $(OUT)/target
-	@$(VECHO) "#define TARGET_RISCV 1" > $@
-	@ln -s $(PWD)/$(SRCDIR)/riscv-codegen.c $(SRCDIR)/codegen.c
+# set ARM by default
+ifeq ($(strip $(ARCH)),riscv)
+ARCH = riscv
 else
-	@$(VECHO) "$(ARM_EXEC)" > $(OUT)/target
-	@$(VECHO) "#define TARGET_ARM 1" > $@
-	@ln -s $(PWD)/$(SRCDIR)/arm-codegen.c $(SRCDIR)/codegen.c
+ARCH = arm
 endif
-	@$(VECHO) "Target machine code switch to %s\n" "$$(cat out/target | sed 's/.*qemu-\([^ ]*\).*/\1/')"
+
+TARGET_EXEC = $($(shell echo $(ARCH) | tr a-z A-Z)_EXEC)
+
+config:
+	$(Q)ln -s $(PWD)/$(SRCDIR)/$(ARCH)-codegen.c $(SRCDIR)/codegen.c
+	$(call $(ARCH)-specific-defs) > $@
+	$(VECHO) "Target machine code switch to %s\n" $(ARCH)
 
 $(OUT)/tests/%.elf: tests/%.c $(OUT)/$(STAGE0)
 	$(VECHO) "  SHECC\t$@\n"
diff --git a/mk/arm.mk b/mk/arm.mk
index 0195288..5dbddd3 100644
--- a/mk/arm.mk
+++ b/mk/arm.mk
[...]

PR #49, opened 2021-02-05
Support define symbol as literal

### LINKED ISSUE #48: Can't define symbol as string
Current `shecc` can't define symbol as string, as a result, #47 fail to refactor.

```cpp
#define ARCH_PREDEFIND "__arm__"
```
Above expression is  invalid, later patch will fix it.

eecheng87:
Close #48

### DIFF
--- a/src/cfront.c
+++ b/src/cfront.c
@@ -23,6 +23,21 @@ int is_hex(char c)
             (c >= 'A' && c <= 'F'));
 }
 
+int is_numeric(char buffer[])
+{
+    int i = 0;
+    while (buffer[i]) {
+        if (i == 1 && (buffer[i] == 'x')) {
+            if (buffer[0] != '0')
+                return 0;
+        } else if (is_digit(buffer[i]) == 0) {
+            return 0;
+        }
+        i++;
+    }
+    return 1;
+}
+
 /* lexer tokens */
 typedef enum {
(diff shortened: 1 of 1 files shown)


eecheng87:
One of reason fail to bootstrap in #47 was fixed in this patch. 
It seems that there still are some bugs in #47, I'll figure out later.

src/defs.h
```diff
@@ -154,6 +154,7 @@ typedef struct {
 typedef struct {
     char alias[MAX_VAR_LEN];
     char value[MAX_VAR_LEN];
+    int numeric;
```
jserv, line 157: I just wonder why `numeric` is introduced. Can you eliminate it?

eecheng87: Before this patch, `shecc` always views defined symbol as T_numeric. There's chance be T_string such as `#define a "text"`.

jserv: I see. However, we can still distinguish from string literals on demand.

eecheng87: Yes, we can. If we don't want `numeric` as flag, we need to implement another function to distinguish `char*` from literal or numeric. Is that OK?

jserv: > Yes, we can. If we don't want `numeric` as flag, we need to implement another function to distinguish `char*` from literal or numeric. Is that OK?

It sounds better.

PR #51, opened 2021-02-06
Support function pointer

jserv:
This patch introduced indirect function call, which is required by
function pointers. New opcode "OP_indirect" appears in IR, and both Arm
and RISC-V implementations are implemented accordingly.

Later, we can even encapsulate the target-specific code generations with
function pointers, that means no duplicated "codegen.c" file.


jserv:
Test procedure:
```shell
make clean config ARCH=riscv check
make clean config ARCH=arm check
```
Both are working to me.

Known issue: failed to bootstrap:
```
  SHECC	out/shecc-stage2.elf
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
Segmentation fault (core dumped)
```

jserv:
@eecheng87, Finally, it can self-host and pass the test suite. Ready for reviewing.

eecheng87 (approved):

PR #56, opened 2022-07-31
Allow riscv-codegen to generate executable mul instruction without RV32M

oucs638:
Add new compiler option +m/-m as the switch to take whether m extension or not.
Check the global variable "use_m_ext" before riscv-codegen going to call `__mul()/__div()/__mod().`
If there is no m extension, replace `__mul()/__div()/__mod()` with `__addi(zero, zero, 0)` as the no operation in riscv.


eecheng87:
Please remove generated file `mul_test`

tests/mul_test.c
```diff
@@ -0,0 +1,9 @@
+#include <stdio.h>
```
eecheng87, line 1: What is the difference between this file and driver.sh:L324-331? Can we merge this test suite into driver.sh? I suppose there is something duplicate.

jserv:
@oucs638, would you rework this patch?

jserv:
The proposed change was far away from the mainline. Drop it.

PR #57, opened 2023-03-08
Fix pointer dereferencing

vacantron:


tests/driver.sh
```diff
@@ -204,6 +204,8 @@ EOF
 # pointers
 items 3 "int x; int *y; x = 3; y = &x; return y[0];"
 items 5 "int b; int *a; b = 10; a = &b; a[0] = 5; return b;"
+items 2 "int x[2]; int y; x[1] = 2; y = *(x + 1); return y;"
+items 2 "int x; int y; int *z; z = 2; y = &z; x = *y; return x;"
```
eecheng87, line 208: I think "x = *y" might lead to syntax error, did it?

vacantron: No. The error mentioned in section of **Known Issues** only occurs when attempting to assign value to a dereferenced pointer. Reading from reference won't cause this.

eecheng87: The second test case you gave might lead to syntax error (invalid type argument of unary '*' for `y`). Maybe you intend to do this: `int x; int *y; int z; z = 2; y = &z; x = *y; return x;`

Issue #43, opened 2021-01-20
Didn't support "#ifdef ... #else ... #endif"


eecheng87:
As title, current `shecc` didn't support "#ifdef ... #else ... #endif".

PR #61, opened 2023-07-22
Implement preprocessor directive ifdef-else-endif

ChinYikMing:
#43

### DIFF
--- a/tests/driver.sh
+++ b/tests/driver.sh
@@ -323,4 +323,20 @@ int main()
 }
 EOF
 
+# #ifdef...#else...#endif
+try_ 0 << EOF
+#define A 0
+#define B 200
+int main()
+{
+    int x;
+#ifdef A
+    x = A;
+#else
+    x = B;
+#endif
+    return x;
+}
+EOF
+
 echo OK
--- a/src/cfront.c
+++ b/src/cfront.c
@@ -151,19 +151,34 @@ token_t get_next_token()
             return T_define;
         }
         if (!strcmp(token_str, "#ifdef")) {
+            char *alias = NULL;
             i = 0;
             do {
                 token_str[i++] = next_char;
             } while (read_char(0) != '\n');
             token_str[i] = 0;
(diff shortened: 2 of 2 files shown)


jserv (changes requested):
Append `Close #43` at the end of git commit message.

jserv:
Check [CONTRIBUTING.md](https://github.com/jserv/shecc/blob/master/CONTRIBUTING.md) to enforce consistent style.

ChinYikMing:
passed the checks

tests/driver.sh
```diff
@@ -323,4 +323,20 @@ int main()
...
+#define B 200
+int main()
+{
+    int x;
+#ifdef A
+    x = A;
+#else
+    x = B;
+#endif
+    return x;
```
jserv, line 338: The return value is crucial for automated tests. Ensure that 0 is returned in the expected path.

jserv (changes requested):
The string `Close #43` should appear in the body of git commit message rather than the subject. In addition, read https://cbea.ms/git-commit/ carefully to consolidate the commit message.

jserv:
I defer to @eecheng87 to validate the proposed changes.

eecheng87 (approved):

jserv:
You may consider to implement `#if defined(...)` and `#elif`, so that we can refine the existing code such as [lib/c](https://github.com/jserv/shecc/blob/master/lib/c.c). Expected usage:
```c
#if defined(__arm__)
#define __syscall_exit 1
#define __syscall_read 3
#define __syscall_write 4
#define __syscall_close 6
#define __syscall_open 5
#define __syscall_brk 45

#elif defined(__riscv)
#define __syscall_exit 93
#define __syscall_read 63
#define __syscall_write 64
#define __syscall_close 57
#define __syscall_open 1024
#define __syscall_openat 56
#define __syscall_brk 214

#else
#error Unsupported architecture
#endif
```

PR #62, opened 2023-07-25
Implement #if defined(...) and #elif defined(...) and #else

ChinYikMing:
Now only support simple conditional expression which is defined(...) after #if and #elif. More complex conditional expression might be supported in the future.

Refactor 'lib/c.c' to improve readability.

Also, add a simple test case to 'tests/driver.sh'


jserv:
Can you reuse the logic between `#ifdef` and `#if defined`?

ChinYikMing:
Do you mean use `#ifdef` and `#if defined` at the same time?

jserv:
> Do you mean use `#ifdef` and `#if defined` at the same time?

I meant, to avoid code duplication for the same functionalities mentioned above.

ChinYikMing:
I think some logic of `#if` and `#elif` can be refactored. Is it possible for next commit?

jserv:
> I think some logic of `#if` and `#elif` can be refactored. Is it possible for next commit?

You can submit multiple functional commits in this pull request.

jserv:

PR #63, opened 2023-07-25
Eliminate overflow in format manipulation

Alx-Lai:
There's an overflow in `__format()` and `__str_base10` function causing printing error in `printf()` function.
Below are the results run `tests/test.c`

before:
```
F(10) = 2147483647
F(10) = -
F(10) = 2147483647
F(10) = -214748364
F(10) =  -214748364
F(10) =       -214748364
```

after(expected):
```
F(10) = 2147483647
F(10) = -2147483648
F(10) = -2147483647
F(10) = -214748364
F(10) =  -214748364
F(10) =       -214748364
```


jserv (changes requested):
Run `clang-format -i` to enforce consistent style.

Alx-Lai:
try run following test code like amacc do
```python
import subprocess as sp
proc = sp.run(['qemu-arm', '<the binary>'], timeout=10, stdout=sp.PIPE)
print(proc.stdout, proc.stderr, proc.returncode)
```
but got
```bash
$ python3 test.py
2147483647b'' None 0
```

which shows failed to get `qemu-arm`'s output via `subprocess` module

Also,
```bash
x=$(qemu-arm <the binary>); echo "1" $x
```
got
```
<binary output>1
```
which indicates failed to get the output too.

however, with `strace -f qemu-arm <binary>`
```
...
[pid  6712] write(0, "2147483647", 102147483647)  = 10
...
```
it output to fd 0 (stdout)

Is there any way to fix this problem?

jserv (changes requested):

Alx-Lai:
Find out to write to fd 0 (stdin) in origin code. Thus I cannot get output from either stdout or stderr.......

jserv:
Read [How to Write a Git Commit Message](https://cbea.ms/git-commit/) and rewrite git commit messages accordingly.

jserv (changes requested):
Be aware of the rules while writing git commit messages.
* [Capitalize the subject line](https://cbea.ms/git-commit/#capitalize)
* [Wrap the body at 72 characters](https://cbea.ms/git-commit/#wrap-72)
* [Use the body to explain what and why vs. how](https://cbea.ms/git-commit/#why-not-how)

Use `git rebase -i` to rework the commits.

tests/driver.sh
```diff
@@ -357,4 +379,47 @@ int main()
 }
 EOF
 
+# format
+try_output 0 "2147483647" << EOF
```
jserv, line 383: I wonder if here `0` is necessary or not. Can we eliminate it?

Alx-Lai: Probably not. It is rare to have a test program to check both exit code and output. I'll remove it.

Alx-Lai: How about providing both expected exit code and expected output and integrate them into function `try`.

jserv: > How about providing both expected exit code and expected output and integrate them into function `try`.

I agree with this proposal. Please go ahead.

tests/driver.sh
```diff
@@ -6,58 +6,47 @@ readonly SHECC="$PWD/out/shecc"
 
 function try() {
```
jserv, line 7: Add comment as the usage prior to function body `try()`.

tests/driver.sh
```diff
@@ -6,58 +6,47 @@ readonly SHECC="$PWD/out/shecc"
...
-        exit 1
-    fi
+    try "$expected" "$expected_output" "$input"
 }
 
 function items() {
     local expected="$1"
     local input="$2"
-    try "$expected" "int main(int argc, int argv) { $input }"
+    try "$expected" "" "int main(int argc, int argv) { $input }"
```
jserv, line 49: It is not elegant. Instead, we can obtain the number of arguments passed to a shell function.
See https://linuxhint.com/check-the-number-of-arguments-in-bash/

tests/driver.sh
```diff
@@ -132,7 +156,7 @@ items 30 "int i; int acc; i = 0; acc = 0; do { i = i + 1; if (i - 1 < 5) continu
 items 26 "int acc; acc = 0; int i; for (i = 0; i < 100; i++) { if (i < 5) continue; if (i == 9) break; acc = acc + i; } return acc;"
 
 # functions
-try_ 55 << EOF
+try_ 55 "" << EOF
```
jserv, line 159: Can you avoid `""` here?

Alx-Lai: I think there are two cases to call `try_` function
1. compare both exit code and output
2. compare exit code only
Thus, there are two solutions.
1. Another argument check in `try_`
2. write another testing function like `try_`

jserv:
Thank @Alx-Lai for contributing!

PR #65, opened 2023-07-27
Support conversion specifier "%c" inside printf

### LINKED ISSUE #64: Support conversion specifier “%c” inside printf
It would be good if we can print out and see the character at some points especially on debugging. Also, it is an improvement of  the functionality of printf function.

@jserv @eecheng87 May I push a new commit?

ChinYikMing:
Close #64


jserv:

PR #67, opened 2023-07-30
Improve format specifiers

Alx-Lai:
In padding section of format function, use to pad less than 100(10^2) character. And add support of "%%" format, which simply print '%' character.

### DIFF
--- a/lib/c.c
+++ b/lib/c.c
@@ -21,7 +21,7 @@
 
 #endif
 
-#define INT_BUF_LEN 11
+#define INT_BUF_LEN 16
 
 typedef int FILE;
 
@@ -95,7 +95,7 @@ void __str_base10(char *pb, int val)
     int neg = 0;
 
     if (val == -2147483648) {
-        strncpy(pb, "-2147483648", 11);
+        strncpy(pb + INT_BUF_LEN - 11, "-2147483648", 11);
         return;
     }
     if (val < 0) {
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Enforce the term used in C standard library: format specifiers.
See [printf(3)](https://man7.org/linux/man-pages/man3/printf.3.html).

lib/c.c
```diff
@@ -153,9 +153,21 @@ void __str_base10(char *pb, int val)
     }
 }
 
+
```
jserv, line 156: Avoid unnecessary blank lines.

lib/c.c
```diff
@@ -174,19 +186,26 @@ int __format(char *buffer,
...
     char pb[INT_BUF_LEN];
     int pbi = 0;
 
-    if (hexprefix == 1) {
-        buffer[0] = '0';
-        buffer[1] = 'x';
-        bi = 2;
-        if (width > 2)
+    if (alternate_form == 1) {
+        if (base == 8) {
```
jserv, line 196: Replace `if` and `else if` statement with `switch`.

lib/c.c
```diff
@@ -198,12 +217,14 @@ int __format(char *buffer,
 
     if (base == 10)
```
jserv, line 218: Ditto. Use `switch` statement.

jserv:
Thank @Alx-Lai for contributing!

PR #68, opened 2023-08-06
Support function-like macro

vacantron:
The implementation redirects the pointer to the source to the beginning of the body of the macro, like how the copy and paste works. Because of this, the answer of the second test is 7, not 9.

The `__VA_ARGS__` identifier will expand the remained arguments that are undeclared. It is useful to write the more readable codes.

**Not** support nested macro yet.

### DIFF
--- a/src/cfront.c
+++ b/src/cfront.c
@@ -1,8 +1,17 @@
 /* C language front-end */
 
+char peek_char(int offset);
 int is_whitespace(char c)
 {
-    return (c == ' ' || c == '\r' || c == '\n' || c == '\t');
+    if (c == '\\' && peek_char(1) == '\n')
+        return 2;
+
+    return (c == ' ' || c == '\t');
+}
+
+int is_newline(char c)
+{
+    return (c == '\r' || c == '\n');
 }
 
 /* is it alphabet, number or '_'? */
@@ -111,13 +120,26 @@ typedef enum {
 char token_str[MAX_TOKEN_LEN];
 token_t next_token;
 char next_char;
+int skip_newline = 1;
 
 int preproc_match;
+int macro_return_idx;
 
 void skip_whitespace()
 {
-    while (is_whitespace(next_char))
-        next_char = SOURCE[++source_idx];
+    while (1) {
+        if (is_whitespace(next_char) == 2) {
+            source_idx += 2;
+            next_char = SOURCE[source_idx];
+            continue;
+        }
+        if (is_whitespace(next_char) ||
+            (skip_newline && is_newline(next_char))) {
+            next_char = SOURCE[++source_idx];
+            continue;
+        }
+        break;
+    }
 }
 
 char read_char(int is_skip_space)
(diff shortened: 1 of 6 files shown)


src/defs.h
```diff
@@ -9,7 +9,7 @@
 #define MAX_PARAMS 8
 #define MAX_LOCALS 48
 #define MAX_FIELDS 32
-#define MAX_FUNCS 1024
+#define MAX_FUNCS 2048
```
jserv, line 12: Can you lower?

vacantron: Sure. The `func_tries_idx` almost reaches this limit before changing, so I just double the value of it.

Actually, the `funcs_idx` only uses low hundreds of it. I add another limitation to separate these two.

src/cfront.c
```diff
@@ -1,8 +1,17 @@
 /* C language front-end */
 
+char peek_char(int offset);
 int is_whitespace(char c)
 {
-    return (c == ' ' || c == '\r' || c == '\n' || c == '\t');
+    if (c == '\\' && peek_char(1) == '\n')
+        return 2;
```
jserv, line 7: Don't do that. `is_whitespace` is supposed to be returning `0` or `1`.

src/cfront.c
```diff
@@ -678,7 +722,7 @@ void lex_expect(token_t token)
     next_token = get_next_token();
 }
 
-void read_expr(int param_no, block_t *parent);
+void read_expr(int param_no, block_t *parent, macro_t *macro);
```
jserv, line 725: Don't do that. Avoid changing the function prototypes. The macros should be regarded as a special case of expressions.

vacantron: We need to check the variables are live in that block or not. The macro is in the same way, but it doesn't need to know how the variable is exactly. The only thing it should know is where the argument from (the position in the source code).

In practice, the arguments of the macro have the higher priority to search that cause the behavior different from the normal variables. We search in the scope of the macro first, then the block. Or, we can say the scope of a macro is the **superset** of the block it belonging.

jserv: > Or, we can say the scope of a macro is the **superset** of the block it belonging.

I agree with this concept. Let's unify within the pull request.

src/cfront.c
```diff
@@ -1,8 +1,19 @@
 /* C language front-end */
 
+char peek_char(int offset);
```
jserv, line 3: Move the forward declaration `peek_char` right before  the `is_line_continuation`.

src/cfront.c
```diff
@@ -1,8 +1,19 @@
...
 int is_whitespace(char c)
 {
-    return (c == ' ' || c == '\r' || c == '\n' || c == '\t');
+    return (c == ' ' || c == '\t');
+}
+
+int is_line_continuation(char c)
```
jserv, line 9: Would it be better to call `is_line_break`?

jserv (changes requested):
Update `README.md` for improving the features described in section "Compatibility."

src/globals.c
```diff
@@ -126,12 +129,13 @@ ir_instr_t *add_instr(opcode_t op)
     return ii;
 }
 
-block_t *add_block(block_t *parent, func_t *func)
+block_t *add_block(block_t *parent, func_t *func, macro_t *macro)
```
vacantron, line 132: This change is to indicate that the block can access variables from function and both macro parameter declarations. I think it makes the semantics clearer than passing the pointer of macro from parent.

jserv: Describe the design decisions in git commit messages as well.

src/cfront.c
```diff
@@ -111,13 +122,26 @@ typedef enum {
 char token_str[MAX_TOKEN_LEN];
 token_t next_token;
 char next_char;
+int skip_newline = 1;
 
 int preproc_match;
+int macro_return_idx;
```
jserv, line 128: Add comments for macro implementation.

jserv:

PR #70, opened 2023-08-08
Wrap verbose IR generation via function-like macro

vacantron:
1. Use macro to perform the conditional compilation, and dump the IR in the debug mode only.
2. Modify the frontend for parsing the function-like macro without body.
3. Update makefiles:
    - The previous `Makefile` does not check the `ARCH` that user typing in. If it is neither `arm` nor `riscv`, the exception will arise. Fix this by checking the argument first.
    - Make the template for auto-generated configure file prettier.

### DIFF
--- a/mk/arm.mk
+++ b/mk/arm.mk
@@ -9,8 +9,8 @@ export ARM_EXEC
 
 arm-specific-defs = \
     $(Q)$(PRINTF) \
-" /* target: ARM */\n\
-  \#define ARCH_PREDEFINED \"__arm__\" /* defined by GNU C and RealView */\n\
-  \#define ELF_MACHINE 0x28 /* up to ARMv7/Aarch32 */\n\
-  \#define ELF_FLAGS 0x5000200\n\
-"
+        "/* target: ARM */\n$\
+        \#define ARCH_PREDEFINED \"__arm__\" /* defined by GNU C and RealView */\n$\
+        \#define ELF_MACHINE 0x28 /* up to ARMv7/Aarch32 */\n$\
+        \#define ELF_FLAGS 0x5000200\n$\
+        "
--- a/mk/arm.mk
+++ b/mk/arm.mk
@@ -9,8 +9,8 @@ export ARM_EXEC
 
 arm-specific-defs = \
     $(Q)$(PRINTF) \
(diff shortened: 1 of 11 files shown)


mk/arm.mk
```diff
@@ -9,8 +9,8 @@ export ARM_EXEC
 
 arm-specific-defs = \
     $(Q)$(PRINTF) \
-" /* target: ARM */\n\
-  \#define ARCH_PREDEFINED \"__arm__\" /* defined by GNU C and RealView */\n\
-  \#define ELF_MACHINE 0x28 /* up to ARMv7/Aarch32 */\n\
-  \#define ELF_FLAGS 0x5000200\n\
-"
+        "/* target: ARM */\n$\
```
jserv, line 12: Don't do this! Minimize the necessary changes.

vacantron: This makes the generated configure file aligned left. Note there's a whitespace at the beginning of lines before.
```diff
- /* target: ARM */
- #define ARCH_PREDEFINED "__arm__" /* defined by GNU C and RealView */
- #define ELF_MACHINE 0x28 /* up to ARMv7/Aarch32 */
- #define ELF_FLAGS 0x5000200
+/* target: ARM */
+#define ARCH_PREDEFINED "__arm__" /* defined by GNU C and RealView */
+#define ELF_MACHINE 0x28 /* up to ARMv7/Aarch32 */
+#define ELF_FLAGS 0x5000200
```
Replacing `$\` with `\` and rearranging the string to eliminate that and keep the content easy to read.

src/defs.h
```diff
@@ -29,6 +29,23 @@
 #define ELF_START 0x10000
 #define PTR_SIZE 4
 
+#if defined(DEBUG)
```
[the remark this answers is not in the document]
jserv: I tend to introduce `DUMP_IR` macro and drop the use of `DEBUG` macro, resulting in minimal changes to the build system.

mk/arm.mk
jserv: > Replacing `$\` with `\` and rearranging the string to eliminate that and keep the content easy to read.

Do that in another pull request.

jserv:

PR #71, opened 2023-08-09
Enhance argument passing for syscall

WangHanChi:
In the previous version, only three arguments were passed for system calls. Therefore, I expanded it to support the passing of up to six arguments for system calls like mmap.


vacantron (changes requested):
Please leave a blank line between the title and the description of the commit message.

jserv:
Thank @WangHanChi for contributing! The git commit message was amended, and it does make sense to introduce register allocation proposed by @vacantron for better code generation.

PR #72, opened 2023-08-09
Avoid variable conflicts while expanding macros

vacantron:
Rename the counter `i` in `DUMP_IR` macro to prevent the conflict after macro expansion.

Fix the dumped IR format.

### DIFF
--- a/src/defs.h
+++ b/src/defs.h
@@ -29,17 +29,18 @@
 #define ELF_START 0x10000
 #define PTR_SIZE 4
 
-#define DUMP_IR(...)                            \
-    do {                                        \
-        int i;                                  \
-        if (!dump_ir)                           \
-            break;                              \
-                                                \
-        printf("%#010x     ", code_start + pc); \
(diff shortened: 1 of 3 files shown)


src/defs.h
```diff
@@ -29,17 +29,18 @@
...
-            break;                              \
-                                                \
-        printf("%#010x     ", code_start + pc); \
-        for (i = 0; i < _c_block_level; i++)    \
-            printf("    ");                     \
-                                                \
-        printf(__VA_ARGS__);                    \
+#define DUMP_IR(...)                                                 \
+    do {                                                             \
+        int dump_ir_i;                                               \
```
jserv, line 34: Rename from `dump_ir_i` to `__dump_ir_iter`.

vacantron: Should we avoid using the name with double underscore prefix?

jserv: > Should we avoid using the name with double underscore prefix?

It is fine for function-like macros.

PR #73, opened 2023-08-09
Improve architecture detection and configuration

vacantron:
Check the architecture before the building process starting.

Eliminate the leading whitespace in the template string for the auto-generated configure file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ STAGE1 := shecc-stage1.elf
 STAGE2 := shecc-stage2.elf
 
 OUT ?= out
-ARCH := arm
+ARCH ?= arm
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -13,6 +13,7 @@ STAGE1 := shecc-stage1.elf
 STAGE2 := shecc-stage2.elf
 
 OUT ?= out
+ARCH := arm
```
jserv, line 16: Should be `ARCH ?= arm`

vacantron: It has not been set and used before. I think it should be fine?

jserv: No, the variable may be set in automated build environments such as CI pipeline.

PR #74, opened 2023-08-12
Implement #undef preprocessor derictive

ChAoSUnItY:
Implements `#undef` preprocessor derictive. 

One concern in current approach is that due to the memory allocations of `alias_t *ALIASES` and `macro_t *MACROS`, the removal operation's time complexity is `O(N)`, this might causes some performance overhead in future development.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -145,24 +145,57 @@ void add_alias(char *alias, char *value)
     alias_t *al = &ALIASES[aliases_idx++];
     strcpy(al->alias, alias);
     strcpy(al->value, value);
+    al->disabled = 0;
 }
 
 char *find_alias(char alias[])
 {
     int i;
     for (i = 0; i < aliases_idx; i++)
-        if (!strcmp(alias, ALIASES[i].alias))
+        if (!ALIASES[i].disabled && !strcmp(alias, ALIASES[i].alias))
             return ALIASES[i].value;
     return NULL;
+}
+
+int remove_alias(char *alias)
+{
+    int i;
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -6,3 +6,6 @@ a.out
 *.lst
 config
 src/codegen.c
+
+# Vscode
```
jserv, line 10: Don't do this! Minimize the necessary changes.

src/globals.c
```diff
@@ -147,7 +147,7 @@ void add_alias(char *alias, char *value)
     strcpy(al->value, value);
 }
 
-char *find_alias(char alias[])
+char *find_alias(char *alias)
```
jserv, line 150: Don't do this. Minimize the necessary changes.

ChinYikMing:
Maybe You can introduce a `disabled` variable in `alias_t` structure. If the variable is set then it cannot be used otherwise it can. This can improve performance from O(n) to O(1). So, `remove_alias` can be simplified to set the `disabled` variable instead. In addition, you may need to modify the `find_alias` function to consider the `disabled` variable.

But, I think #undef is not used so that much for now, so moving around the element when calling `remove_alias` is acceptable.

ChAoSUnItY:
I see. Should I open another issue for this suggestion?

jserv:
> I see. Should I open another issue for this suggestion?

Simply rework the commits for preprocessor specific changes.

jserv (changes requested):
Squash the commits and improve git commit message.

Issue #20, opened 2020-10-25
libc: Drop brk in favor of mmap inside malloc/free


jserv:
At present, `malloc` was implemented via `brk` system call, which is obsolete. Most malloc implementations use `mmap` system call instead. We shall drop the use of `brk` in favor of `mmap` while implementing `maloc` and `free`.

Check comprehensive list for system calls via https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md#cross_arch-numbers

```c
#define __syscall_mmap 192
#define __syscall_munmap 91
```

PR #75, opened 2023-08-14
Drop brk in favor of mmap inside malloc/free

WangHanChi:
In past implementation, malloc was implemented via brk system call, which is obsolete. Most malloc implementations use mmap system call instead.
Therefore, I warpped up mmap for malloc, and munmap for free.

TODO : Migrate Best-Fit algorithm.

### DIFF
--- a/lib/c.c
+++ b/lib/c.c
@@ -9,6 +9,8 @@
 #define __syscall_close 6
 #define __syscall_open 5
 #define __syscall_brk 45
+#define __syscall_mmap2 192
+#define __syscall_munmap 91
 
 #elif defined(__riscv)
 #define __syscall_exit 93
@@ -18,6 +20,8 @@
 #define __syscall_open 1024
 #define __syscall_openat 56
 #define __syscall_brk 214
+#define __syscall_mmap2 222
+#define __syscall_munmap 215
 
 #endif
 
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Improve git commit message (ChatGPT is your friend) and append "Close #20" at the end of that message.

lib/c.c
```diff
@@ -436,80 +440,94 @@ int fputc(int c, FILE *stream)
     return 0;
 }
 
-/* Reference:
- *   https://danluu.com/malloc-tutorial/
- * FIXME: adopt lite_malloc from musl-libc
- *   https://git.musl-libc.org/cgit/musl/tree/src/malloc/lite_malloc.c
+/*  FIXME: Best-fit Algorithm
+ *   https://github.com/picolibc/picolibc
```
jserv, line 444: Drop this hyperlink. Minimize the necessary changes.

lib/c.c
```diff
@@ -436,80 +440,94 @@ int fputc(int c, FILE *stream)
...
-    block->free = 0;
-    return block;
+    return ((size - 1) | mask) + 1;
 }
 
 void *malloc(int size)
 {
-    block_meta_t *block;
-    if (size == 0)
+    int flags = MAP_PRIVATE | MAP_ANONYMOUS;
```
jserv, line 471: At the moment, there is NO optimizer, which means that additional computation at runtime is required for bitwise-OR and assignment. Thus, you shall pre-calculate these values (with proper comments) and specify them in mmap directly.

lib/c.c
```diff
@@ -436,80 +440,94 @@ int fputc(int c, FILE *stream)
...
+        head->next = NULL;
+        head->prev = NULL;
+        head->ptr = NULL;
+        head->size = 0;
     }
-    return block + 1;
+
+    chunk *cur = head;
+
+    while (cur->next != NULL) {
```
jserv, line 488: Rewrite as following:
```c
    while (cur->next != NULL)
        cur = cur->next;
```

lib/c.c
```diff
@@ -436,80 +440,94 @@ int fputc(int c, FILE *stream)
...
 
-    /* TODO: merge several free memory blocks */
-    block_meta_t *block_ptr = ptr - sizeof(block_meta_t);
-    if (block_ptr->free == 1)
-        abort();
-    block_ptr->free = 1;
-}
+    chunk *cur = head;
+
+    while (cur->ptr != ptr) {
```
jserv, line 511: Ditto. Shorten the loop code.

lib/c.c
```diff
@@ -436,80 +440,89 @@ int fputc(int c, FILE *stream)
     return 0;
 }
 
-/* Reference:
- *   https://danluu.com/malloc-tutorial/
- * FIXME: adopt lite_malloc from musl-libc
- *   https://git.musl-libc.org/cgit/musl/tree/src/malloc/lite_malloc.c
- */
+#define PROT_READ 0x01
```
jserv, line 443: Don't define these. Instead, specify pre-calculated values as the parameters in system calls.

lib/c.c
```diff
@@ -436,80 +440,89 @@ int fputc(int c, FILE *stream)
...
-            if (current->free == 1 && current->size >= size)
-                return current + 1;
-            __malloc_global_last = current;
-            current = current->next;
-        }
-        block = __malloc_request_space(size);
-        if (!block)
-            return NULL;
+
+    /* MAP_PRIVATE | MAP_ANONYMOUS */
```
jserv, line 470: Move the comments inline. That is,
```c
    int flags = 34; /* MAP_PRIVATE (0x02) | MAP_ANONYMOUS (0x20) */
```

lib/c.c
```diff
@@ -436,80 +440,83 @@ int fputc(int c, FILE *stream)
...
 
-/* Reference:
- *   https://danluu.com/malloc-tutorial/
- * FIXME: adopt lite_malloc from musl-libc
- *   https://git.musl-libc.org/cgit/musl/tree/src/malloc/lite_malloc.c
- */
+#define PAGESIZE 4096
 
-typedef struct block_meta {
+typedef struct _chunk {
```
jserv, line 445: Rewrite as following:
```c
typedef struct chunk {
    struct chunk *next;
    struct chunk *prev;
    int size;
    void *ptr;
} chunk_t;
```

PR #76, opened 2023-08-16
Implement #error preprocessor directive

ChAoSUnItY:
Implement `#error` preprocessor directive. The reason not to consume `#error` (`T_error`) token on match is to avoid post side effect of `lex_expect`, in other word, to preseve the correct position of `next_char`.

### DIFF
--- a/src/cfront.c
+++ b/src/cfront.c
@@ -2485,7 +2485,7 @@ void read_global_statement()
 
         remove_alias(alias);
         remove_macro(alias);
-    } else if (lex_peek(T_error, token)) {
+    } else if (lex_peek(T_error, NULL)) {
         int i = 0;
         char error_diagnostic[MAX_LINE_LEN];
 
--- a/src/cfront.c
+++ b/src/cfront.c
@@ -109,6 +109,7 @@ typedef enum {
     T_do,
     T_define,
     T_undef,
+    T_error,
     T_include,
     T_typedef,
     T_enum,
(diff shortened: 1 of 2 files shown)


tests/driver.sh
```diff
@@ -403,6 +403,15 @@ int main()
...
 
+# #error ...
+try_output 1 "" << EOF
+#error This is an error directive
+int main()
+{
+    return 0;
+}
+EOF
+
```
vacantron, line 414:
```
Error This is an error directive at source location 10937, IR index 2236
tests/driver.sh: line 16: Aborted                 "$SHECC" -o "$tmp_exe" "$tmp_in"
#error This is an error directive
int main()
{
    return 0;
}
exit code => 1
output =>
```
@jserv Should we keep this in test? We seem to assume the build should be passed before testing.

jserv: > Should we keep this in test? We seem to assume the build should be passed before testing.

The proposed test case contradicts the assumptions of the current test suite by introducing scenarios that unexpectedly bypass the compilation process. Let's discard it.

src/cfront.c
```diff
@@ -2480,6 +2485,16 @@ void read_global_statement()
 
         remove_alias(alias);
         remove_macro(alias);
+    } else if (lex_peek(T_error, token)) {
```
vacantron, line 2488: The copy of `T_error` seems to be redundant. You can set it to `NULL`.

jserv:

PR #77, opened 2023-08-21
Implement freelist and Best-Fit allocation

WangHanChi:
A simple Freelist and Best-Fit algorithm to reduce frequent use mmap system call. Furthermore, the function for memory deallocation of global objects in shecc has been added, along with the release of all allocated memory chunks upon program termination.


src/main.c
```diff
@@ -69,5 +69,8 @@ int main(int argc, char *argv[])
     /* output code in ELF */
     elf_generate(out);
 
-    return 0;
+    /* release global objects */
```
jserv, line 72: Should be "release allocated objects"

lib/c.c
```diff
@@ -464,37 +468,123 @@ void *malloc(int size)
...
 
-    chunk_t *new =
-        __syscall(__syscall_mmap2, NULL, align_up(sizeof(chunk_t) + size), prot,
-                  flags, -1, 0);
-    cur->next = new;
-    new->prev = cur;
+    chunk_t *best_fit_chunk = NULL;
+    chunk_t *allocated;
+
+    if (!freelist_head->next) {
```
jserv, line 499: Add some comments about best-fit.

Issue #22, opened 2020-10-28
Not enough register for reading expression


eecheng87:
There's a test case:
```
a = 1 + (2 + (3 + (4 + (5 + (6 + (7 + (8 + (9 + (10 + (11 + (12 + (13 + (14 + (15 + (16 + (17 + (18 + (19 + 20))))))))))))))))))
```
In shecc, it will store all constants in register. Obviously, there's not enough register.
In modern compiler, I found they just load an immediate. e.g.
```
li a4, 210
```
Even I mix identifier into expression, modern compiler still load an immediate value. Do we need to rewrite read expression? Maybe we need to evaluate expression before run time. Also, I think we need symbol table to store these identifier.

jserv:
An effective evaluator is required for parsed expression. Which parts of C front-end would be modified to fit?

eecheng87:
Maybe I need to modify `read_expr` which also contains `read_operand` (most complicate part).
It seems modern compiler will also evaluate result of function in expression during compile time, should we follow this? Following is example:

```
int f(int b){
	int c = 10;
	return b * c + 1;
}
void _start(){
    int a;
    int c = 10;
    a = 1 + ( f(1) + (1 + (1 + (1 + (1 + (1 + (1 + (1 + (1 + (1 + (20 + (3 + (1 + (1 + (6 + (1 + (1 + (1 + 20))))))))))))))))));
    print(a);
```
corresponding asm

```
00010068 <_start>:
   10068:	400027b7          	lui	a5,0x40002
   1006c:	04b00713          	li	a4,75
   10070:	00e78023          	sb	a4,0(a5) # 40002000 <__global_pointer$+0x3fff0788>
```

Issue #79, opened 2023-09-07
Improve intermediate representation and also register allocation


vacantron:
## Description
The current register allocator uses the strategy of the stack-based virtual machine and is connected to the generation of the intermediate representation. These might result in the inefficient utilization of the available registers and make the optimization harder.

To improve these, we could decouple the IR generation and the allocator, and introduce the new algorithm (e.g. linear-scan, graph coloring, etc.) to support the register-based virtual machine model.

## Related issues
- #45

PR #80, opened 2023-09-19
Introduce new intermediate representation and register allocator

vacantron:
As #79 mentioned, the stack-based virtual machine uses the available registers inefficiently in RISC-like architecture. The newly two-phase IR replaces the origin one which was coupled with the the register allocation and made the optimization harder.

The first phase IR uses the three-address code (3AC) to simplify the complex expression of C. The register allocator could use these information to generate the more efficient instructions, using the algorithm that supports the register-base virtual machine. It use the linear-scan algorithm in this implementation.

The second phase IR is going to locate the symbol which is used in branch/jump instruction. Once the amount of the other instructions is confirmed, the offset/address would be correct.

The new ARM backend uses register r0 ~ r7 for general purposes, and register r8 as the temporary storage for some instructions. Register r12 is as the global stack pointer for the convenience.

The new RISC-V backend uses register a0 ~ a7 for general purposes, and register t0 as the temporary storage for some instructions.


lib/c.c
```diff
@@ -398,6 +463,7 @@ FILE *fopen(char *filename, char *mode)
 #endif
 
     abort();
+    return -1;
```
jserv, line 466: Is this change necessary?

vacantron: For the present, the function with non-void return type is expected to have a return statement. Maybe we could remove `abort()` and return `NULL`, and then check the `FILE` pointer be valid or not when using?

jserv: Let's drop the call of `abort`.

src/arm-codegen.c
```diff
@@ -1,532 +1,457 @@
 /* Translate IR to target machine code */
 
 #include "arm.c"
+#include "reg-alloc.c"
```
jserv, line 4: Instead of the inclusion of `reg-alloc.c`, can you specify the header?

src/globals.c
```diff
@@ -261,7 +307,7 @@ func_t *find_func(char func_name[])
 {
     int index = find_trie(FUNC_TRIES, func_name);
     if (index)
-        return &FUNCS[index - 1]; /* As the recorded index is a positive
+        return &FUNCS[index]; /* As the recorded index is a positive
```
jserv, line 310: Refine the comment.

src/arm-codegen.c
vacantron: Is it better that moving `dump_ir()`and the inclusion of `reg-alloc.c` into `main.c`?

src/reg-alloc.c
```diff
@@ -0,0 +1,843 @@
+/* Allocate register from IR. Using linear-scan algorithm */
```
jserv, line 1: Mention the limitations as well.

lib/c.c
```diff
@@ -346,23 +346,88 @@ void printf(char *str, ...)
...
+            if (str[si] >= '1' && str[si] <= '9') {
+                w = str[si] - '0';
+                si++;
+                if (str[si] >= '0' && str[si] <= '9') {
+                    w = w * 10;
+                    w += str[si] - '0';
+                    si++;
+                }
+            }
+            if (str[si] == 's') {
```
jserv, line 391: Use switch-case for readable code.

vacantron: May I change it with `printf()` together in another pull request?

src/defs.h
```diff
@@ -104,32 +101,23 @@ typedef enum {
...
     char var_name[MAX_VAR_LEN];
     int is_ptr;
     int is_func;
     int array_size;
-    int offset;   /* offset from stack or frame */
+    int offset;   /* offset from stack or frame, index 0 is reserved */
     int init_val; /* for global initialization */
+
+    int is_global;
+    int eol; /* end-of-life */
```
jserv, line 119: Shall it be named as "liveness"?

src/reg-alloc.c
```diff
@@ -0,0 +1,844 @@
+/**
+ * Allocate register from IR. Using linear-scan algorithm, need at least 7
```
jserv, line 2: Rewrite as following:
> Allocate registers from IR. The linear-scan algorithm now expects a minimum of 7 available registers (typical for RISC-style architectures).

lib/c.c
jserv: For newly-added code, it's better to make improvements sooner rather than requiring extra effort later on.

src/cfront.c
```diff
@@ -835,68 +879,68 @@ void read_inner_var_decl(var_t *vd, int anon)
...
-         * create dummy parameters to put all on stack
-         */
-        for (; vn < MAX_PARAMS; vn++) {
-            strcpy(vds[vn].type_name, "int");
-            strcpy(vds[vn].var_name, "var_arg");
-            vds[vn].is_ptr = 1;
-        }
+        /* variadic function. Max 8 parameters are passed. */
+        fd->va_args = 1;
+        vn = 8;
```
jserv, line 907: Use `MAX_PARAMS` when possible.

jserv:
Is it feasible to close #22 with register allocation?

vacantron:
> Is it feasible to close #22 with register allocation?

Yes. I update the limitation of the operand stack size in [b41e76c](https://github.com/sysprog21/shecc/pull/80/commits/b41e76cf4809f410d8a51d165a045b148f7a0a10) . It can pass the test now.

jserv:
@vacantron, Use `git rebase -i` to rework these git commits, including updating the commit messages, which may include statements like `Close #22`. I am preparing to merge this pull request.

jserv:

PR #81, opened 2023-10-10
Speed up `malloc()`

vacantron:
The `malloc()` in `lib/c.c` is slow. Assume the following code:

```c
#define MAX_ITEMS 100000

typedef struct list_t {
    struct list_t *next;
} list_t;

int main(int argc)
{
    int i;
    list_t *p = malloc(sizeof(list_t));

    for (i = 0; i < MAX_ITEMS; i++) {
        p->next = malloc(sizeof(list_t));
        p = p->next;
    }
    return 0;
}
```

The above code takes 1.5 minutes to execute on QEMU. The reason is that it needs to traverse every chunk until the end of the chunk list. To speed up the tail insertion, we could use a tail pointer or a doubly circular linked list. Considering that there are few bidirectional accesses in memory allocation, the simple tail pointer is used here.

### DIFF
--- a/lib/c.c
+++ b/lib/c.c
@@ -521,6 +521,7 @@ int align_up(int size)
 }
 
 chunk_t *head;
+chunk_t *tail;
 chunk_t *freelist_head;
 
 void *malloc(int size)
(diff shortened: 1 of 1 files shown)


jserv:

PR #85, opened 2023-11-11
Integrate SSA-based middle-end and snapshot tests

vacantron:
The current cfront and IR generation are high coupling that making the flow analysis harder. To introduce the further optimizations, the static single assignment (SSA) is imperative. This patch introduces the minimal changes for it, and also keeps the old first phase IR (ph1_ir) for the compatibility to the upcoming works.

The current unit test only covers the end to end case. It is hard the classify which part went wrong in the compilation. The new middle end can dump the IR with the dot file as a snapshot. We can also use the tool like `graphviz` to check the flow during the compilation.

Use `make check-snapshots` to invoke the snapshot test, and use `tests/update-snapshots.sh` to update the snapshots if there's any change for the cfont.

**Notice:** The snapshots should be updated under the **ARM** configure.


vacantron:
The GitHub Action reports the broken pipe if the snapshot testing failed. It works fine on my local PC and only reports `Files /dev/fd/63 and /dev/fd/62 differ`. Do I mess up something?

https://github.com/vacantron/shecc/actions/runs/6834732304/job/18587784272#step:3:467

jserv:
It is a bit confusing that there are 'check' and 'test' targets which deal with the test suite. Can you clarify?

src/defs.h
```diff
@@ -108,6 +112,26 @@ typedef enum {
 
 /* variable definition */
 typedef struct {
+    int counter;
+    int stack[64];
+    int stack_idx;
+} rename_t;
+
+struct BB;
```
jserv, line 120: To maintain a consistent coding style, I would recommend using `basic_block_t` instead of the overly simplified `BB`.

See https://github.com/icxd/llvm-ir/blob/main/lib/llvm.h

jserv: By the way, it is certainly acceptable to name an instance of `basic_block_t` as `bb` for brevity.

src/globals.c
```diff
@@ -376,6 +379,147 @@ int size_var(var_t *var)
...
+        bb->symbol_list.tail->next = sym;
+        bb->symbol_list.tail = sym;
+    }
+}
+
+/* linked list */
+void add_inst(block_t *block,
+              BB *bb,
+              opcode_t op,
+              var_t *rd,
```
jserv, line 496: Would it be better to reference a structure rather than passing these parameters?

jserv (changes requested):
Rename `src/SSA.c` to `src/ssa.c` (lowercase).
Don't check in JSON files which are generated by snapshot.

jserv:
> The GitHub Action reports the broken pipe if the snapshot testing failed. It works fine on my local PC and only reports `Files /dev/fd/63 and /dev/fd/62 differ`. Do I mess up something?

I got the following messages from CI:
```
int main(int argc, int argv) { int i; int j; i=0; j=0; while (i<10) { j=j+i; i=i+1; } return j; }
exit code => 45
output => 
tests/driver.sh: line 16:  3327 Segmentation fault      (core dumped) "$SHECC" -o "$tmp_exe" "$tmp_in"
int main(int argc, int argv) { int x; x=0; do {x = x + 1; break;} while (1); return x; }
exit code => 1
output => 
tests/driver.sh: line 16:  3334 Aborted                 (core dumped) "$SHECC" -o "$tmp_exe" "$tmp_in"
int main(int argc, int argv) { int x; x=0; do {x++; continue; abort();} while (x < 2); return x; } => 2 expected, but got 1
input: /tmp/tmp.gUdRxRqzMl.c
make: *** [Makefile:54: test] Error 1
```

src/globals.c
vacantron: The old first phase IR will be and replaced by SSA CFG and dropped in the future.

vacantron:
>> https://github.com/vacantron/shecc/actions/runs/6834732304/job/18587784272#step:3:467

I forgot to provide the error messages from GitHub Action:
```shell
tests/check-snapshots.sh
Error: writing output failed: Broken pipe
cat: write error: Broken pipe
make: *** [Makefile:54: check] Error 1
Error: writing output failed: Broken pipe
cat: write error: Broken pipe
Error: Process completed with exit code 1.
```

> I got the following messages from CI:

That message was caused by the mentioned control-flow change. The snapshot testing had been passed. I edited `lib/c.c` and tried to cause the test failed and received the above messages.

jserv (changes requested):
Drop the generated JSON file and update `.gitignore`.

src/defs.h
```diff
@@ -219,3 +254,90 @@ typedef struct {
...
+    int index;
+    struct symbol_t *next;
+} symbol_t;
+
+typedef struct {
+    symbol_t *head;
+    symbol_t *tail;
+} symbol_list_t;
+
+struct basic_block_t {
```
jserv, line 303: Should be `struct basic_block`.

jserv, line 303: Should be `struct basic_block`.

vacantron:
> Drop the generated JSON file and update `.gitignore`.

We need those JSON files as the baseline of tests, right? The generated snapshots by `mktemp` are not in the work directory.

src/defs.h
```diff
@@ -219,3 +261,103 @@ typedef struct {
...
+    int visited;
+    struct basic_block_t *dom_next[64];
+    struct basic_block_t *dom_prev;
+    struct fn_t *belong_to;
+    block_t *scope;
+    symbol_list_t symbol_list; /* variable declaration */
+    int elf_offset;
+};
+
+typedef struct basic_block_t basic_block_t;
```
jserv, line 346: Should be `typedef struct basic_block basic_block_t;`

src/main.c
```diff
@@ -70,20 +73,29 @@ int main(int argc, char *argv[])
...
         dump_ph1_ir();
 
+    build_ssa(dump_ir);
+
+    /* SSA-based liveness analyses */
+    liveness_analysis();
+
     /* allocate register from IR */
-    reg_alloc();
+    allocate();
```
jserv, line 82: Can we keep `reg_alloc` to avoid potential naming violations?

src/main.c
```diff
@@ -70,20 +73,29 @@ int main(int argc, char *argv[])
...
+
+    /* flatten CFG to linear instruction */
+    flatten();
```
jserv, line 85: Ditto. Use better names.

src/main.c
```diff
@@ -70,20 +73,29 @@ int main(int argc, char *argv[])
...
 
     /* dump second phase IR */
     if (dump_ir)
         dump_ph2_ir();
 
     /* generate code from IR */
-    code_generate();
+    generate();
```
jserv, line 92: `generate` is too generic, and we should avoid using it.

src/main.c
```diff
@@ -70,20 +73,29 @@ int main(int argc, char *argv[])
...
 
     /* output code in ELF */
     elf_generate(out);
 
     /* release allocated objects */
+    release_ssa();
```
jserv, line 98: Rename to `ssa_release`.

src/ssa.c
```diff
@@ -0,0 +1,1140 @@
+void forward_bb_traversal(fn_t *fn,
+                          basic_block_t *bb,
+                          void (*preorder_callback)(fn_t *, basic_block_t *),
```
jserv, line 3: Rename to `preorder_cb` and `postorder_cb` where `cb` means callback.

src/ssa.c
```diff
@@ -0,0 +1,1140 @@
...
+                             postorder_callback);
+    if (bb->else_ && bb->else_->visited < fn->visited)
+        forward_bb_traversal(fn, bb->else_, preorder_callback,
+                             postorder_callback);
+
+    if (postorder_callback)
+        postorder_callback(fn, bb);
+}
+
+void backward_bb_traversal(fn_t *fn,
```
jserv, line 25: To avoid potential naming conflicts, I would like to propose consistent naming scheme.
* basic block facilities: starting with `bb_`. e.g., `bb_forward_traversal`
* dominator tree: starting with `dom_`. e.g., `dom_build`.

src/defs.h
```diff
@@ -108,6 +112,26 @@ typedef enum {
...
 /* variable definition */
 typedef struct {
+    int counter;
+    int stack[64];
+    int stack_idx;
+} rename_t;
+
+struct basic_block_t;
+
+struct ref_block_t {
```
jserv, line 122: No suffix `_t` is allowed in structure definition. Instead, suffix `_t` appears in the `typedef`. That is,
```c
struct ref_block {
    basic_block_t *bb;
    struct ref_block *next;
};

typedef struct ref_block ref_block_t;
```

src/arm-codegen.c
```diff
@@ -445,3 +445,358 @@ void code_generate()
...
+
+    int i;
+    for (i = 0; i < ph2_ir_idx; i++) {
+        ph2_ir_t *ph2_ir = &PH2_IR[i];
+        int rd = ph2_ir->dest;
+        int rn = ph2_ir->src0;
+        int rm = ph2_ir->src1;
+        int ofs;
+
+        switch (ph2_ir->op) {
```
jserv, line 621: Can you move some parts of this function out? It is lengthy.

src/defs.h
```diff
@@ -174,15 +210,21 @@ typedef struct {
...
+struct ph2_ir_t {
     opcode_t op;
     int src0;
     int src1;
     int dest;
     char func_name[MAX_VAR_LEN];
     char true_label[MAX_VAR_LEN];
     char false_label[MAX_VAR_LEN];
-} ph2_ir_t;
+    struct basic_block_t *next_bb;
```
jserv, line 221: Enforce use of `typedef`. That is, prefer `basic_block_t *next_bb` over `struct basic_block *bb`.

src/arm-codegen.c
vacantron: Using the compiler directive could extract this part from both ARM and RISC-V backend, but the readability might be worse. Is it recommended?

jserv: No, simply use function calls to split the code.

src/cfront.c
```diff
@@ -2519,7 +2754,17 @@ void read_body_statement(block_t *parent)
...
+
+        basic_block_t *body_ = create_BB(parent);
+        connect_BB(cond_, body_, THEN);
+        body_ = read_body_statement(parent, body_);
+
+        if (body_) {
+            connect_BB(body_, inc_, NEXT);
+            connect_BB(inc_, cond_, NEXT);
+        } else {
+            /* release dangling inc basicblock */;
```
jserv, line 2766: Add FIXME or TODO in comments.

src/globals.c
```diff
@@ -376,6 +381,152 @@ int size_var(var_t *var)
     return s;
 }
 
+/* linked list */
```
jserv, line 384: The comment is not meaningful comparing to the function name.

src/globals.c
```diff
@@ -376,6 +381,152 @@ int size_var(var_t *var)
...
+        FUNC_LIST.head = n;
+        FUNC_LIST.tail = n;
+        return n;
+    }
+    FUNC_LIST.tail->next = n;
+    FUNC_LIST.tail = n;
+    return n;
+}
+
+basic_block_t *create_BB(block_t *parent)
```
jserv, line 399: Enforce the consistent coding style.

README.md
```diff
@@ -105,6 +110,11 @@ $ qemu-arm fib
 $ make check
 ```
 
+Also, check the emitted IRs are identical with the snapshots with "check-snapshots" argument:
```
jserv, line 113: Proofread as following:
> Also, verify that the emitted IRs are identical to the snapshots by specifying `check-snapshots` target when invoking `make`.

src/cfront.c
```diff
@@ -1010,29 +1016,34 @@ void read_func_parameters(block_t *parent)
     for (i = 0; i < param_num; i++) {
         ph1_ir_t *ph1_ir = add_ph1_ir(OP_push);
         ph1_ir->src0 = params[i];
+        /* Liveness extension in allocation uses `param_num - i` */
```
jserv, line 1019: More explanations here.

src/cfront.c
```diff
@@ -1693,6 +1763,7 @@ void read_lvalue(lvalue_t *lvalue,
     }
 }
 
+/* add_inst: change pended */
```
jserv, line 1766: I didn't get it. Improve the comment.

vacantron: Some changes are still continuing, check [5106846](https://github.com/sysprog21/shecc/pull/85/commits/5106846760c656f87fbfdec955b0467826e65293). I'll complete them soon.

vacantron:
`shecc` now passes all the unit tests before, but the cfront lacks the support for some syntax, likes structure declaration without `typedef`. So the self-hosting still fails now.

src/arm-codegen.c
```diff
@@ -7,157 +7,183 @@ void emit(int code)
...
+    func->fn->bbs->elf_offset = 44;
 
-    add_label("__syscall", 44);
-
-    /* calculate the offset of labels */
-    elf_code_idx = 96;
-
-    for (i = 0; i < ph2_ir_idx; i++) {
-        ph2_ir = &PH2_IR[i];
+    elf_offset = 80;
```
jserv, line 15: Add comments here.

ChAoSUnItY (approved):

src/defs.h
```diff
@@ -131,25 +160,30 @@ typedef struct {
     int disabled;
 } macro_t;
 
+typedef struct fn fn_t;
```
jserv, line 163: It is confusing to have two derived types, `fn_t` and `func_t`. Can you clarify?

jserv: `typedef` is useful if we want to emphasize the types. However, it may become redundant once it is referred to be a shortcut of structures.

jserv:
> `shecc` now passes all the unit tests before, but the cfront lacks the support for some syntax, likes structure declaration without `typedef`. So the self-hosting still fails now.
> The instructions which loads/stores with a huge offset also need to be modified when self-hosting.

I plan to merge this pull request soon so that we can proceed with the refinement of the C front-end or the implementation of additional optimization phases. Can you make the minimal changes necessary to achieve this goal?

src/defs.h
vacantron: > It is confusing to have two derived types, fn_t and func_t. Can you clarify?

Those two one should be the same structure, but the difference here is that one is declared to a static array and the other is dynamically allocated. I keep the old one for the compatibility to @ChAoSUnItY 's works.

vacantron:
> > `shecc` now passes all the unit tests before, but the cfront lacks the support for some syntax, likes structure declaration without `typedef`. So the self-hosting still fails now.
> > The instructions which loads/stores with a huge offset also need to be modified when self-hosting.
> 
> I plan to merge this pull request soon so that we can proceed with the refinement of the C front-end or the implementation of additional optimization phases. Can you make the minimal changes necessary to achieve this goal?

I think it's already the minimal changes. I have avoided to change the existing workflow and only appended the support for the SSA. The old first phase IR works fine also (tho it should be dropped in the future).

jserv:
> I think it's already the minimal changes. I have avoided to change the existing workflow and only appended the support for the SSA. The old first phase IR works fine also (tho it should be dropped in the future).

Change the state from Draft when you are ready to merge. In addition, address the action items in git commit messages.

src/arm-codegen.c
```diff
@@ -7,157 +7,184 @@ void emit(int code)
...
-            elf_code_idx += 24;
-            break;
-        case OP_log_and:
-            elf_code_idx += 28;
+        case OP_global_store:
+            elf_offset += 4;
             break;
         default:
-            break;
+            printf("Unknown OP\n");
```
jserv, line 29: Replace with the message "Unknown opcode."

src/arm-codegen.c
```diff
@@ -7,157 +7,184 @@ void emit(int code)
...
+                    else
+                        elf_offset += 4;
+
+                    break;
+                case OP_store:
+                case OP_global_store:
+                    flatten_ir = add_ph2_ir(OP_address_of);
+                    memcpy(flatten_ir, inst, sizeof(ph2_ir_t));
+
+                    if (flatten_ir->src1 > 255)
```
jserv, line 100: Add comments to explain the load/store model.

src/arm-codegen.c
```diff
@@ -166,40 +193,51 @@ void code_generate()
...
+            emit(__add_i(__AL, rd, __r12, ph2_ir->src0));
+            break;
+        case OP_assign:
+            emit(__mov_r(__AL, rd, rn));
+            break;
+        case OP_global_store:
+            emit(__sw(__AL, rn, __r12, ph2_ir->src1));
+            break;
+        default:
+            printf("Unknown OP\n");
```
jserv, line 219: Ditto.

src/arm-codegen.c
```diff
@@ -411,37 +406,23 @@ void code_generate()
...
-            emit(__mov_i(__EQ, rd, 1));
-            break;
-        case OP_rshift:
-            emit(__srl(__AL, rd, rn, rm));
-            break;
-        case OP_lshift:
-            emit(__sll(__AL, rd, rn, rm));
-            break;
         default:
+            printf("Unknown OP\n");
```
jserv, line 424: Ditto.

src/defs.h
jserv: > Those two ones would be the same structure, but the difference here is one is declared to a static array and the other is dynamically allocated. I keep the old one for the compatibility to @ChAoSUnItY 's works.

Add comments with corresponding FIXME/TODO notation.

src/globals.c
```diff
@@ -376,6 +381,154 @@ int size_var(var_t *var)
...
+
+    if (str)
+        strcpy(n->str, str);
+
+    if (!bb->inst_list.head) {
+        bb->inst_list.head = n;
+        bb->inst_list.tail = n;
+    } else {
+        bb->inst_list.tail->next = n;
+        bb->inst_list.tail = n;
```
jserv, line 528: Move the statement `bb->inst_list.tail = n;` out of if-else blocks.

src/riscv-codegen.c
```diff
@@ -7,151 +7,149 @@ void emit(int code)
...
+                    break;
+                case OP_return:
+                    flatten_ir = add_ph2_ir(OP_return);
+                    memcpy(flatten_ir, inst, sizeof(ph2_ir_t));
+                    /* restore sp */
+                    flatten_ir->src1 = bb->belong_to->func->stack_size;
+                    elf_offset += 16;
+                    break;
+                default:
+                    printf("Unknown OP\n");
```
jserv, line 128: Ditto.

src/riscv-codegen.c
```diff
@@ -395,32 +340,21 @@ void code_generate()
...
-            emit(__xori(rd, rd, 1));
-            break;
-        case OP_rshift:
-            emit(__sra(rd, rs1, rs2));
-            break;
-        case OP_lshift:
-            emit(__sll(rd, rs1, rs2));
-            break;
         default:
+            printf("Unknown OP\n");
```
jserv, line 356: Ditto.

src/ssa.c
```diff
@@ -0,0 +1,1152 @@
+void bb_forward_traversal(fn_t *fn,
```
jserv, line 1: How about defining a structure which contains `fn`, `bb`, and two callbacks? Therefore, we can operate with an explicit pointer without passing these parameters.

src/ssa.c
```diff
@@ -0,0 +1,1152 @@
...
+    while (i != j) {
+        while (i->rpo > j->rpo)
+            i = i->idom;
+        while (j->rpo > i->rpo)
+            j = j->idom;
+    }
+    return i;
+}
+
+void build_idom()
```
jserv, line 102: Add comments to address the use of dominator tree.

src/ssa.c
```diff
@@ -0,0 +1,1152 @@
...
+            break;
+        }
+
+    if (found)
+        return;
+
+    bb->live_kill[bb->live_kill_idx++] = var;
+}
+
+void add_killed_block(basic_block_t *bb, var_t *var)
```
jserv, line 236: Can you unify the naming scheme? That is, use prefix `bb_`.

src/ssa.c
```diff
@@ -0,0 +1,1152 @@
...
+            case OP_leq:
+            case OP_bit_and:
+            case OP_bit_or:
+            case OP_bit_xor:
+            case OP_log_and:
+            case OP_log_or:
+                sprintf(
+                    str,
+                    "<%s<SUB>%d</SUB> := %s<SUB>%d</SUB> %s %s<SUB>%d</SUB>>",
+                    inst->rd->var_name, inst->rd->subscript,
```
jserv, line 776: Can you replace the lengthy parameter with a function call?

src/ssa.c
vacantron: That function is for `var` to record which block kill itself. Is it better to swap the arguments?

jserv: Up to you. I just concern the naming scheme, avoiding potential conflicts.

src/arm-codegen.c
vacantron: I misunderstood the instruction in ARMv7-A manual. The immediate value uses 12 digits instead of 8 digits. I'll fix it.

jserv:
Did you drop snapshot tests? If so, change the subject of this pull request.

src/defs.h
```diff
@@ -219,3 +256,119 @@ typedef struct {
...
+    opcode_t opcode;
+    var_t *rd;
+    var_t *rs1;
+    var_t *rs2;
+    int sz;
+    phi_operand_t *phi_ops;
+    char str[64];
+};
+
+typedef struct Inst Inst_t;
```
jserv, line 280: Enforce snakecase and use "insn" to represent "instruction." That is, we get `insn_t`.

jserv (changes requested):
Let's unify our terminology. From now on, use `insn` as the abbreviation for "Instruction," rather than "inst."

src/ssa.c
```diff
@@ -0,0 +1,1249 @@
...
+
+        bb->prev[i].bb = NULL;
+    }
+    free(bb);
+}
+
+void ssa_release()
+{
+    fn_t *fn;
+    bb_traversal_args_t *args = calloc(1, sizeof(bb_traversal_args_t));
```
jserv, line 1240: When did you release `args`?

src/defs.h
vacantron: > Let's unify our terminology. From now on, use insn as the abbreviation for "Instruction," rather than "inst."

So the "inst_t" is going to be "insn_t", right?

jserv: > So the "inst_t" is going to be "insn_t", right?

Exactly.

PR #89, opened 2023-12-10
Revise for self-hosting at commit 2b5d7b6

vacantron:
To process the ~700 lines inline libc, the below macro is set:

```diff
+ #define MAX_ANALYSIS_STACK_SIZE 700
```

Now, `shecc` can bootstrap itself at the older (2b5d7b6) version.


src/arm-codegen.c
```diff
@@ -27,18 +27,39 @@ void cfg_flatten()
...
+                elf_offset += 4;
+            break;
         default:
             printf("Unknown opcode\n");
             abort();
         }
     }
-    /* jump to main */
-    elf_offset += 4;
+    /* prepare argc, argv and jump to main */
```
jserv, line 61: Refine the grammar.

src/arm-codegen.c
```diff
@@ -226,23 +247,51 @@ void code_generate()
...
+            } else
+                emit(__sw(__AL, rn, __r12, ph2_ir->src1));
             break;
         default:
             printf("Unknown opcode\n");
             abort();
         }
     }
-    /* jump to main */
+    /* prepare argc, argv and jump to main */
```
jserv, line 289: Ditto.

src/defs.h
```diff
@@ -37,6 +37,7 @@
 #define MAX_CASES 128
 #define MAX_NESTING 128
 #define MAX_OPERAND_STACK_SIZE 32
+#define MAX_ANALYSIS_STACK_SIZE 700
```
jserv, line 40: How can you measure and/or calculate the value?

vacantron: Loading an address from .data section increments the stack index of `live_kill` by 1 in analysis. The inline libc has ~700 times loading, so the limitation is set to 700.

src/arm-codegen.c
vacantron: They are "prepare argc, argv" and "jump to main". Are they too confusing?

src/defs.h
jserv: You should explain within git commit message(s).

src/arm-codegen.c
jserv: Yes, use Grammarly or similar tools to refine.

src/arm-codegen.c
```diff
@@ -27,18 +27,39 @@ void cfg_flatten()
          ph2_ir = ph2_ir->next) {
         switch (ph2_ir->op) {
         case OP_load_constant:
+            if (ph2_ir->src0 < 0)
```
jserv, line 30: Shall we introduce a helper function for the adjust of `elf_offset`?

src/riscv-codegen.c
```diff
@@ -54,107 +119,259 @@ void cfg_flatten()
...
         }
     }
 }
 
+void emit(int code)
+{
+    elf_write_code_int(code);
+}
+
+void emit_from_ph2_ir(ph2_ir_t *ph2_ir)
```
jserv, line 147: Use shorter form. e.g., `emit_ph2ir`.

jserv:

PR #92, opened 2023-12-14
Separate lexer and parser from cfront

ChAoSUnItY:
This pull request separates the lexer and parser into independent modules from original cfront module. This is the predecessor task for later tokenizer optimization (e.g. less side effect).


src/lexer.c
```diff
@@ -0,0 +1,743 @@
...
+    return ((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || c == 'x' ||
+            (c >= 'A' && c <= 'F'));
+}
+
+int is_numeric(char buffer[])
+{
+    int i, hex, size = strlen(buffer);
+
+    if (size > 2)
+        hex = (buffer[0] == '0' && buffer[1] == 'x') ? 1 : 0;
```
jserv, line 128: Rewrite as following:
```c
int hex = 0;
if (size > 2)
    hex = (buffer[0] == '0' && buffer[1] == 'x') ? 1 : 0;
```

No `else` statement is required.

jserv:
I defer to @vacantron for reviewing.

jserv:

Issue #93, opened 2023-12-16
High branch-miss rate when hosting shecc on the Raspberry Pi 3B


vacantron:
The built executable file (here is shecc itself at 2b5d7b6) has much higher branch-miss rate (19.91%) compare with GCC built (5.74%) when running on the Raspberry Pi 3B. The branch misprediction might potentially slowdown the execution.

## How to reproduce

1. Build 2b5d7b6
```shell
$ make config
$ make out/shecc
$ out/shecc -o out/shecc-stage1.elf ../shecc-2b5d7b6/src/main.c
$ chmod +x out/shecc-stage1.elf
```
2. Measure with perf
```shell
$ perf stat -r 5 -e branches,branch-misses out/shecc-stage1.elf tests/fib.c
```
3. Get the statistics
```shell
 Performance counter stats for 'out/shecc-stage1.elf tests/fib.c' (5 runs):

         2,308,457      branches:u                #   33.270 M/sec                    ( +-  0.00% )
           459,550      branch-misses:u           #   19.91% of all branches          ( +-  0.02% )

         0.078757 +- 0.000616 seconds time elapsed  ( +-  0.78% )
```
## Related issues

- #86 
- #88

PR #95, opened 2023-12-16
Fix branch-misses on Raspberry Pi 3B

vacantron:
According to ["Arm Cortex-A53 MPCore Processor Technical Reference Manual"](https://developer.arm.com/documentation/ddi0500/j/Level-1-Memory-System/L1-Instruction-memory-system/Program-flow-prediction), the data-processing instructions using the PC as a destination register is not predicted.

After substituting the `MOV pc, ...` with `BLX` instruction, the statistics become:
```shell
Performance counter stats for 'out/shecc-stage1.elf -o e.out tests/fib.c' (5 runs):

         2,308,525      branches:u                                                    ( +-  0.00% )
           229,797      branch-misses:u           #    9.95% of all branches          ( +-  0.09% )

         0.0756673 +- 0.0000838 seconds time elapsed  ( +-  0.11% )
```

It reduces half of branch-misses and has ~4% speedup. The remaining part should be improved in the optimizer.

## Related Issues
-  #93

### DIFF
--- a/src/arm-codegen.c
+++ b/src/arm-codegen.c
@@ -315,7 +315,7 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
         emit(__movt(__AL, __r8, ph2_ir->src1 + 4));
         emit(__add_r(__AL, __sp, __sp, __r8));
         emit(__lw(__AL, __lr, __sp, -4));
-        emit(__mov_r(__AL, __pc, __lr));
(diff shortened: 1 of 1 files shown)


jserv:
Should we close #93 accordingly?

vacantron:
> Should we close #93 accordingly?

OK. I have update the commit message.

PR #96, opened 2023-12-19
Eliminate compiler warnings

vacantron:
The cfront does not support casting yet. Skip it using the preprocessing.


src/defs.h
```diff
@@ -45,6 +45,16 @@
...
 #define REG_CNT 8
 
+/* This macro will be automatically defined at shecc run-time. */
+#ifdef __SHECC__
+#define UNUSED(x) (x)
+#define SHECC_CAST(x) (x)
+#else
+/* suppress GCC/Clang warnings */
+#define UNUSED(x) (void) (x)
+#define SHECC_CAST(x) (void *) (x)
```
jserv, line 55: I don't get the idea why `SHECC_CAST` exists. Can you explain?

jserv: Is it necessary?

vacantron: > I don't get the idea why `SHECC_CAST` exists. Can you explain?

In printf/fprintf, GCC/Clang will check the format string specified with `-Wall` flag
```c
fprintf(fd, "subgraph cluster_%p {\n", SHECC_CAST(bb));
```

> Is it necessary?

Or we could simply apply the `-Wno-format` flag.

jserv: > Or we could simply apply the `-Wno-format` flag.

Let's enforce `-Wno-format` option.

PR #97, opened 2023-12-22
Revise for self-hosting

vacantron:
- add `calloc()`
- add `HOST_PTR_SIZE` macro for memory copying
- support structure definition
- revise predefined limitations
- revise `UNUSED()` macro
- revise typedef parsing for structure definition
- fix incorrect `tail` of chunck after `free()`
- swap out unsupported statements to equivalent alternatives
- omit dumping basic blocks during bootstrapping
- update snapshots

## Known issues:
- `free()` takes lots of time that increases run time from 1.1s to 5.4s.


jserv (changes requested):
Explicitly mention forward declaration and current workaround.

src/defs.h
```diff
@@ -48,13 +48,22 @@
 /* This macro will be automatically defined at shecc run-time. */
 #ifdef __SHECC__
 #define UNUSED(x) (x)
+#define HOST_PTR_SIZE 4
 #else
 /* suppress GCC/Clang warnings */
 #define UNUSED(x) (void) (x)
+/* configure host data model when using `memcpy()` */
+#define HOST_PTR_SIZE 8
```
jserv, line 56: Is better to use `UINT_MAX` (defined in `<stdint.h>`) instead?

src/ssa.c
```diff
@@ -680,12 +697,17 @@ void unwind_phi()
...
 
         fn->visited++;
-        args->preorder_cb = &bb_unwind_phi;
+        args->preorder_cb = bb_unwind_phi;
         bb_forward_traversal(args);
     }
     free(args);
 }
 
+/* `cfonrt` does not support the string literal addressing yet. Skip them with
```
jserv, line 706: Rephrase as following:
> The current Cfonrt does not yet support string literal addressing, which results in the omission of basic block visualization during the stage-1 and stage-2 bootstrapping phases.

lib/c.c
```diff
@@ -677,9 +686,15 @@ void free(void *ptr)
     if (!ptr)
         return;
 
+    /* TODO: it takes long time to search in chuncks */
```
jserv, line 689: FYI. An excellent video: [malloc sucks so i wrote my own heap in C](https://youtu.be/CulF4YQt6zA).

src/defs.h
vacantron: > Is better to use `UINT_MAX` (defined in `<stdint.h>`) instead?

We need to configure it according to the host data model. Is it better to use `__SIZEOF_POINTER__` predefined macro?

jserv: For stage-0 bootstrapping, any C99-compatible compilers should be able to compile shecc from scratch. I prefer the macros defined in `<stdint.h>`.

vacantron: > Is better to use `UINT_MAX` (defined in `<stdint.h>`) instead?

The `UINT_MAX` seems not to be used for getting the platform's pointer size.

PR #101, opened 2024-01-03
Refine Makefile on error handling for ARM32 architecture detection

idoleat:
With the original Makefile, the host without `arch` in PATH gets "make: arch: No such file or directory" printed everytime while making, regardless which target is choosed. The `2>/dev/null` is added for suppressing the error message.


jserv:
Thank @idoleat for contribution!

PR #102, opened 2024-01-04
Report some context of an error

redhog:
Figuring out line numbers is hard, but printing the next N characters after the error position is easy, and is usually enough to find where it is using grep :)


jserv (changes requested):
Read https://cbea.ms/git-commit/ and [CONTRIBUTING.md](https://github.com/sysprog21/shecc/blob/master/CONTRIBUTING.md) carefully. Then, refine your code and git commit messages accordingly.

PR #105, opened 2024-01-15
Removes redundant block from parser.c

cratuki:


jserv:
Thank @cratuki for contributing!

Issue #84, opened 2023-11-11
Thoughts on cfront's potential improvements


ChAoSUnItY:
Currently cfront is using scanless parser with IR emitter binds into it, and it contains ~3000 LOC. But based on my contributions experience to industrial grade programming languages (V Lang in this case), shecc's frontend parser is lack of ease to either debug or for others to contribute (Even though shecc is meant to be educational).

Here's a list of my thoughts on improving the frontend of shecc:
1. Discard scanerless parser, rewrite it into both lexer and parser.
2. Introducing Abstract Syntax Tree for better IR emission and backend generation.
3. Separation on the compilation phases into multiple files would be a better idea for potential contributors to learn shecc's architecture.

And by accepting the suggestion described above, the possible major changes would be:
1. Having more than 3 phases compilation in shecc.
2. The code generation logic must rewrite based on the introduced AST.

jserv:
> Currently cfront is using scanless parser with IR emitter binds into it, and it contains ~3000 LOC. But based on my contributions experience to industrial grade programming languages (V Lang in this case), shecc's frontend parser is lack of ease to either debug or for others to contribute (Even though shecc is meant to be educational).

I have undertaken the task of reworking my earlier compiler project, [AMaCC](https://github.com/jserv/amacc), with a primary focus on its educational value and potential extensibility. I wholeheartedly acknowledge the limitations of the current C front-end implementation, including the absence of a robust AST and proper modularization.

At the same time, @vacantron is dedicated to enhancing the SSA based IR following his initial efforts in register allocation. I want to make sure that we avoid any significant conflicts of interest when it comes to the reworking of the C front-end. Could you please consider proposing a plan for submitting pull requests that involve minimal changes?

jserv:
> Currently cfront is using scanless parser with IR emitter binds into it, and it contains ~3000 LOC. But based on my contributions experience to industrial grade programming languages (V Lang in this case), shecc's frontend parser is lack of ease to either debug or for others to contribute (Even though shecc is meant to be educational).

I have undertaken the task of reworking my earlier compiler project, [AMaCC](https://github.com/jserv/amacc), with a primary focus on its educational value and potential extensibility. I wholeheartedly acknowledge the limitations of the current C front-end implementation, including the absence of a robust AST and proper modularization.

[...]

PR #106, opened 2024-01-16
Migrate preprocessor directive handling

ChAoSUnItY:
This pull request migrates preprocessor directive handling to parser unit, which unify the behaviour of lexer unit by only handling lexical analysis. This also improves preprocessor directive parsing by processing it with token instead of raw string.

Notice that in this pull request, there are several hacks were introduced due to previous faulty design (See `lexer.c#lex_accept`).

This pull request is part of  #84.

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -58,10 +58,6 @@ typedef enum {
     T_while,
     T_for,
     T_do,
-    T_define,
-    T_undef,
-    T_error,
-    T_include,
     T_typedef,
     T_enum,
     T_struct,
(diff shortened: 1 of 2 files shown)


src/lexer.c
```diff
@@ -71,7 +67,16 @@ typedef enum {
     T_case,
     T_break,
     T_default,
-    T_continue
+    T_continue,
+    T_preproc_include,
```
jserv, line 71: Use `cppd` to stand for C PreProcessor Directives. That is, `T_cppd_include`, `T_cppd_define`, etc.

jserv:

PR #107, opened 2024-01-16
Enhance and cleanup lexer-parser interface

ChAoSUnItY:
This pull request cotains the following major changes:

- Separate lex operations so developer can controls the behaviour of preprocessor aliasing, this also replaces `preproc_aliasing` functionality.
- Refine control flow related preprocessor directives functions: Migrate `if_elif_skip_lines` and `ifdef_else_skip_lines` into single function due to the nature of same functionality and covers all possiblities of control flow combines. (In previous design, former one does not cover `#else` and later one does not cover `#elif`.)
- Refine naming of lex related functions.

In addition, this pull request fixes hacks introduced in #106.

This pull request is part of #84.

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -174,7 +174,9 @@ char peek_char(int offset)
     return SOURCE[source_idx + offset];
 }
 
-/* Lex next token and returns its token type. */
+/* Lex next token and returns its token type. Parameter `aliasing` is used for
+ * disable preprocessor aliasing on identifier tokens.
+ */
 token_t lex_token_internal(int aliasing)
 {
     token_str[0] = 0;
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -87,11 +87,6 @@ int skip_newline = 1;
 
 int preproc_match;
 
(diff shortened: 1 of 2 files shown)


src/lexer.c
```diff
@@ -174,44 +169,13 @@ char read_char(int is_skip_space)
...
-        alias = find_alias(alias_name);
-    } else
-        alias = find_alias(token_str);
-
-    if (alias)
-        preproc_match = 1;
-}
-
-token_t get_next_token()
+/* Lex next token and returns its token type. */
```
jserv, line 177: Improve the comment for describing parameter `aliasing`.

src/parser.c
```diff
@@ -66,6 +66,9 @@ int get_size(var_t *var, type_t *type)
  * whitespace */
 void skip_line(int invalidate)
 {
+    /* FIXME: this invalidation is too aggressive, comment should not be
```
jserv, line 69: Fix grammar mistake.

jserv:

PR #109, opened 2024-02-16
Report additional context when error panicking

ChAoSUnItY:
In previous design, calling `error(char *msg)` in `globals.c` will only show context of error and `source_idx`, which is hard for both developer and contributor to locate the source of error.

In this PR, error message will also show the 1 line of source code and the exact character position where the error occurs.   

```diff
  Error ERR_TEST at source location 100
+ #include <stdlib.h>
+    ^ Error occurs here
```

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -636,8 +636,27 @@ void global_release()
 
 void error(char *msg)
 {
-    /* TODO: figure out the corresponding C source and report line number */
-    printf("Error %s at source location %d\n", msg, source_idx);
+    /* Construct error source diagnostic */
+    int offset, start_idx, i = 0;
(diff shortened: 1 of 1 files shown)


src/globals.c
```diff
@@ -636,8 +636,30 @@ void global_release()
 
 void error(char *msg)
 {
-    /* TODO: figure out the corresponding C source and report line number */
-    printf("Error %s at source location %d\n", msg, source_idx);
+    /* Construct error source diagnostic */
```
jserv, line 639: Proofread:
> Construct error source diagnostics, enabling precise identification of syntax and logic issues within the code.

jserv:

PR #110, opened 2024-02-21
Implement parser for `#include` preprocessor directive

ChAoSUnItY:
- Implement actual parser for `#include` preprocessor directive to avoid potential syntax error.
  - While the parser is implemented, the actual "including" functionality is still implemented in other place.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -62,20 +62,6 @@ int get_size(var_t *var, type_t *type)
     return type->size;
 }
 
-/* abort when invalidate is true and the line contains  character other than
- * whitespace */
-void skip_line(int invalidate)
-{
-    /* FIXME: Comments will causes current validation failed. */
-    skip_whitespace();
-    do {
-        if (invalidate && !is_whitespace(peek_char(0)) &&
-            !is_newline(peek_char(0))) {
-            error("Expects whitespace after preprocessor directive");
-        }
-    } while (read_char(0) != '\n');
-}
-
 /* Skips lines where preprocessor match is false, this will stop once next
  * token is either `T_cppd_elif`, `T_cppd_else` or `cppd_endif`.
  */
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Use `git rebase -i` to squash the commits. Make the git commit messages informative.

jserv:
Can you decouple "Implement parser for #include preprocessor directive & constant evaluation for #if and #elif" into two pull requests? So that, it is more straightforward for reviewing.

src/parser.c
```diff
@@ -62,18 +62,284 @@ int get_size(var_t *var, type_t *type)
...
+    /* Evaluate unary expression first */
+    op = get_operator();
+    current_precedence = get_unary_operator_prio(op);
+    if (current_precedence != 0 && current_precedence >= precedence) {
+        lhs = read_constant_infix_expr(current_precedence);
+
+        switch (op) {
+        case OP_add:
+            break;
+        case OP_sub: {
```
vacantron, line 227: Are the braces `{` `}` in the case statements necessary? Can you remove them?

tests/driver.sh
```diff
@@ -533,4 +533,18 @@ int main()
 }
 EOF
 
+try_ 1 << EOF
+#if -1
+#define A 1
```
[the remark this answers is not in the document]
ChAoSUnItY: My bad, there should be no same value for different branch. Furthermore, returning 1 is not ideal since compilation error will also returns 1 as result, will fix this in later commit.

The patch will be committed in #111.

src/parser.c
ChAoSUnItY: There are unnecessary, I will remove them in later commit.

The patch will be committed in #111.

jserv (changes requested):
Improve the git commit message. You should address the limitation of the proposed change.

jserv (changes requested):
1. Don't indent git commit messages. For example, instead of writing "- This syntax parser can parse 2 forms," simply write "This syntax parser can parse 2 forms." without the leading `- `.
2. Avoid using [backticks](https://en.wikipedia.org/wiki/Backtick) in git commit messages. Instead, use (double) quotation marks. This approach ensures compatibility with most editors.

jserv:

PR #113, opened 2024-02-28
Remove redundant condition

zoanana990:
(c >= '0' && c <= '9') ? 1 : 0;

the condition is redundant

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -119,7 +119,7 @@ int is_alnum(char c)
 
 int is_digit(char c)
 {
-    return (c >= '0' && c <= '9') ? 1 : 0;
+    return (c >= '0' && c <= '9');
 }
 
 int is_hex(char c)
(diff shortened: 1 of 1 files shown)


src/lexer.c
```diff
@@ -119,7 +119,7 @@ int is_alnum(char c)
 
 int is_digit(char c)
 {
-    return (c >= '0' && c <= '9') ? 1 : 0;
+    return (c >= '0' && c <= '9');
```
ChAoSUnItY, line 122: Remove parentheses as well since they does not effect expression precedence.

ChAoSUnItY (changes requested):

src/lexer.c
zoanana990: Okay, I will modify that function and the function like that, is_whitespace and is_newline

zoanana990: I push a new commit c818075e8ff887b81777387805245d294199626b to remove the redundant parentheses.

ChAoSUnItY (approved):

jserv:
Thank @zoanana990 for contributing!

PR #114, opened 2024-02-29
Fix unsafely CSE for global variable

vacantron:
The global variable might be changed by the subroutine. Considering the following code:

```c
int e = arr[idx];
subroutine_changing_idx();
int d = arr[idx];
```

We cannot invoke the CSE and replace `d = arr[idx]` with `d = e` safely here.

### DIFF
--- a/src/ssa.c
+++ b/src/ssa.c
@@ -1067,6 +1067,9 @@ int cse(insn_t *insn, basic_block_t *bb)
         return 0;
 
     var_t *def = NULL, *base = prev->rs1, *idx = prev->rs2;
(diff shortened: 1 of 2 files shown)


src/ssa.c
```diff
@@ -1067,6 +1067,11 @@ int cse(insn_t *insn, basic_block_t *bb)
         return 0;
 
     var_t *def = NULL, *base = prev->rs1, *idx = prev->rs2;
+    if (base->is_global)
```
jserv, line 1070: Rewrite as following:
```c
if (base->is_global || idx->is_global)
    return 0;
```

PR #116, opened 2024-03-03
Implement constant optimization

vacantron:
Consider the following scenario:

```c
int a = 2;            /* constant assingment */
int b = a;            /* assignment via constant representation */
int c = a + b;
int d = c + 8;        /* mixed assigment */
return a + b + c + d; /* chained assignment */
```

The current code generator emits the assembly below, even though the variables a, b, c, d are all constant representations.

```assembly
mov     r0, #2
mov     r1, r0
mov     r0, r1
add     r3, r1, r0
mov     r2, #8
add     r2, r3, r2
add     r4, r1, r0
add     r0, r4, r3
add     r1, r0, r2
mov     r0, r1
```

The new constant optimizer will evaluate the constant representations, trim the unused variables, and emit `mov r0, #20` only.

### DIFF
--- a/src/defs.h
+++ b/src/defs.h
@@ -16,8 +16,8 @@
 #define MAX_PARAMS 8
 #define MAX_LOCALS 1450
 #define MAX_FIELDS 32
-#define MAX_FUNCS 256
-#define MAX_FUNC_TRIES 1950
+#define MAX_FUNCS 512
+#define MAX_FUNC_TRIES 2160
 #define MAX_BLOCKS 1150
 #define MAX_TYPES 64
 #define MAX_IR_INSTR 36864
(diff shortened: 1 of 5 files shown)


jserv (changes requested):
Add related test cases to verify the effectiveness of the proposed optimizer.

vacantron:
> Add related test cases to verify the effectiveness of the proposed optimizer.

The test cases are implicitly included in the existent ones, e.g. https://github.com/sysprog21/shecc/blob/393b441cb271c5b3e500b2d5de518de64fce6328/tests/driver.sh#L81-L83

Should we add more?

jserv:
> The test cases are implicitly included in the existent ones, e.g.
> https://github.com/sysprog21/shecc/blob/393b441cb271c5b3e500b2d5de518de64fce6328/tests/driver.sh#L81-L83
> Should we add more?

Yes, append the new items which reflect the chained assignments.

jserv:

PR #117, opened 2024-03-04
Support variable declaration in for loop setup

### LINKED ISSUE #115: Declare variables where needed
In C99, it is possible to declare variables precisely where they are needed, a feature also present in C++. Variables can also be declared within the control structure of a `for` loop, a feature introduced in C99:
```c
 for (int x = 0; x < 10; x++) {  /* This is permitted starting with C99 */
    ...
}
```

The C99 standard says: (6.8.5.3 The for statement)
> for ( clause-1 ; expression-2 ; expression-3 ) statement
> behaves as follows: The expression expression-2 is the controlling expression that is evaluated before each execution of the loop body. The expression expression-3 is evaluated as a void expression after each execution of the loop body. If clause-1 is a declaration, the scope of any variables it declares is the remainder of the declaration and the entire loop, including the other two expressions; it is reached in the order of execution before the first evaluation of the controlling expression. If clause-1 is an expression, it is evaluated as a void expression before the first evaluation of the controlling expression.

This enhancement simplifies the code by allowing variable declarations closer to where they are used, thereby improving readability and maintainability. shecc should support this feature.

ChAoSUnItY:
- Resolve #115 , this allows variables declare in the scope of for loop body, to avoid loose variable liveness, in other word, variable can have stricter scope liveness by declaring in setup section of for loop.


jserv:
CI complains below:
```
tests/check-snapshots.sh
Error: writing output failed: Broken pipe
cat: write error: Broken pipe
```

@vacantron, can you help resolve?

tests/driver.sh
```diff
@@ -566,4 +566,16 @@ int main()
 }
 EOF
 
+# for-loop var decl
```
jserv, line 569: Change the comment to "Variables can be declared within a for-loop iteration"

src/parser.c
```diff
@@ -2326,16 +2326,63 @@ basic_block_t *read_body_statement(block_t *parent, basic_block_t *bb)
...
         if (!lex_accept(T_semicolon)) {
-            lex_peek(T_identifier, token);
-            read_body_assignment(token, parent, OP_generic, &bb);
+            if (!lex_peek(T_identifier, token))
+                error("Unexpected token");
+
+            int find_type_flag = lex_accept(T_struct) ? 2 : 1;
+            type = find_type(token, find_type_flag);
+            if (type) {
+                var = require_var(parent);
```
vacantron, line 2339: The scope of the declaration in the first clause should be different from the others that outside of the for-statement. i.e. you need to create another scope with `read_code_block()` or something else first.

vacantron (changes requested):

vacantron:
> CI complains below:
> 
> ```
> tests/check-snapshots.sh
> Files /dev/fd/63 and /dev/fd/62 differ
> Error: writing output failed: Broken pipe
> cat: write error: Broken pipe
> Files /dev/fd/63 and /dev/fd/62 differ
> ```
> 
> @vacantron, can you help resolve?

The new cfront feature is going to change the CFGs. Use `tests/update-snapshots.sh` to update the outdated ones.

By the way, the changes causes the unexpected behavior while unwinding the phi instruction in the second stage, and breaks the self-hosting. I'm trying to figure it out.

vacantron:
Rebase the `master` branch after #113 being merged, and the error should be solved.

ChAoSUnItY:
That's weird, I updated snapshot after rebase and it still panics on stage 2.

vacantron:
> That's weird, I updated snapshot after rebase and it still panics on stage 2.

The second stage is broken by the SSA now, but the snapshot test should be passed tho.

I run `head -c20 tests/snapshots/xxx.json | tail -c19` for the `hello` and `fib`, and both of them has 469 basic blocks. It is definitely incorrect. The `fib` one should have ten more than the `hello`. Can you try it again with `make disclean` and `make config ARCH=arm` ?

ChAoSUnItY:
> Can you try it again with `make disclean` and `make config ARCH=arm` ?

After running `disclean` and `config ARCH=arm`, now running `tests/update-snapshots.sh` would stuck forever (I terminted it after running for 2.6 mins). Can you confirm this weird behavior?

vacantron:
> After running `disclean` and `config ARCH=arm`, now running `tests/update-snapshots.sh` would stuck forever (I terminted it after running for 2.6 mins). Can you confirm this weird behavior?

There is a typo in the commands that I provided previously. Try `make distclean config ARCH=riscv out/shecc-stage1.elf` and `tests/update-snapshots.sh` again. Do you run them on Linux?

ChAoSUnItY:
> Do you run them on Linux?
Yes, specifically, on Arch Linux.

> Try make distclean config ARCH=riscv out/shecc-stage1.elf and tests/update-snapshots.sh again.
It still failed on stage 2.

vacantron:
> Yes, specifically, on Arch Linux.

Can you test the `jq` command with `-S` option separately on your machine? The result seems not to be sorted in ascending order by the `_gvid`.

ChAoSUnItY:
I can confirm that using `jq -S -c . hello.json > hello.json` (from `tests/update-snapshot.sh`) does only sorted by json keys but `_gvid` value.

jserv (changes requested):
Rebase the latest `master` branch and re-generate snapshots.

jserv:
> Leave a blank line between the subject and the body of the commit message, and use `git rebase -i` to squash some commits

See also: https://cbea.ms/git-commit/

vacantron (approved):

jserv:

PR #119, opened 2024-03-19
Fix snapshot test

vacantron:
Related PR:
- #117


jserv (changes requested):
Make the commit message informative.

tests/check-snapshots.sh
```diff
@@ -12,7 +12,12 @@ function check_snapshot() {
 
     $SHECC --dump-ir -o $temp_exe $source &>/dev/null
     dot -Tdot_json -o $temp_json CFG.dot
-    diff -q <(cat $ref) <(sed -E "/0x[0-9a-f]+/d" $temp_json | jq -S -c .)
+    diff -q <(cat $ref) \
```
jserv, line 15: I wonder why the proposed change looks much longer.

vacantron: The current snapshot is a JSON object dumped by the `dot` utility, and this JSON object describes the whole CFG:

```json
{
  "_subgraph_cnt": <# of blocks>,
  "directed": true,
  "edges": [
    {
        "_gvid": 0,
        "head": ...,
        "tail": ...,
    },
    ...
  ],
  "label": "",
  "name": "CFG",
  "objects": [
    {
        "_gvid": 0,
        "edges": [
            0,
            1,
            2,
            ...
        ],
        "nodes": [
            475,
            476,
            477,
            ...
        ],
        "subgraphs": [
            1,
            2,
            3,
           ...
        ]
    },
    ...
  ]
}
```

The previous `jq` command didn't sort the keys of the object and the elements of the arrays, and make the `diff` command misjudgment the similarity of them.

jserv: Mention the above in git commit message as well.

PR #124, opened 2024-03-26
Implement macros defined in `<stdbool.h>`

### LINKED ISSUE #122: Support macros defined in <stdbool.h>
C99 introduces a built-in type called `_Bool` for handling Boolean values. Additionally, the standard header `<stdbool.h>` defines macros `bool`, `true`, and `false` to facilitate boolean operations. 

As per C99's specifications, except for bit-fields, all data types, including bool, must be allocated at least one byte of memory. This allocation ensures that objects consist of contiguous byte sequences, the specifics of which depend on the implementation or are explicitly stated. Since a `char`'s size is universally one byte, it follows logically that the size of a `bool` must also be one byte, at most. In summary, based on C99 standards, the size of a `bool` is definitively set to one byte.

Incorporating the `bool` type into shecc would enhance the expressiveness of source code and improve memory usage when expressions solely need to represent binary relationships.

C99 6.2.6.1 General
> Except for bit-fields, objects are composed of contiguous sequences of one or more bytes, the number, order, and encoding of which are either explicitly specified or implementation-defined.

C99 6.3.1.1 Boolean, characters, and integers
> The rank of `_Bool` shall be less than the rank of all other standard integer types.

ChAoSUnItY:
- Resolve #122, implements all `stdbool.h` macros in `lib/c.c`
  - This includes macro `bool`, `true` and `false`.
- Partially utilize `stdbool.h` macros in `lexer.c` and `parser.c`

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -83,16 +83,16 @@ typedef enum {
 char token_str[MAX_TOKEN_LEN];
 token_t next_token;
 char next_char;
-int skip_newline = 1;
+bool skip_newline = true;
 
-int preproc_match;
+bool preproc_match;
 
 /* Point to the first character after where the macro has been called. It is
  * needed when returning from the macro body.
  */
 int macro_return_idx;
 
-int is_whitespace(char c)
+bool is_whitespace(char c)
 {
     return c == ' ' || c == '\t';
 }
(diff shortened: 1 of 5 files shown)


src/lexer.c
```diff
@@ -83,70 +83,70 @@ typedef enum {
...
 }
 
-int is_numeric(char buffer[])
+bool is_numeric(char buffer[])
 {
     int hex = 0, size = strlen(buffer);
 
     if (size > 2)
-        hex = (buffer[0] == '0' && buffer[1] == 'x') ? 1 : 0;
+        hex = buffer[0] == '0' && buffer[1] == 'x';
```
jserv, line 136: Should variable `hex` be declared as `bool`?

jserv:
Can the proposed change deal with `sizeof(_Bool)`?

ChAoSUnItY:
> Can the proposed change deal with `sizeof(_Bool)`?

Now I supported `_Bool` type as builtin type, however, this would cause further issue in `lexer.c` since macro `#define bool _Bool` has rhs token with token type `T_identifier` but previous algorithm only recognize `T_numeric` and `T_string`, I would like to resolve this issue in another PR.

src/parser.c
```diff
@@ -2872,6 +2875,10 @@ void parse_internal()
     type->base_type = TYPE_int;
     type->size = 4;
 
+    type = add_named_type("_Bool");
```
jserv, line 2878: Here, add comments to explain the `_Bool` type as defined in the C99 specification, for developers who may not be familiar with it.

jserv:

Issue #46, opened 2021-02-04
Support Arm/RISC-V targets without integer division instructions


jserv:
Some Arm targets such as Cortex-A9 might lack of integer division instructions, and the current `shecc` would not generate the correct instruction sequence for them. Consequently, we have to provide the fallback by means of software emulation. Here is the reference implementation for integer division:
```c
__div:
        push  {fp, lr}
        add   fp, sp, #4
        // check for divide by zero
        cmp   r1, #0
        beq   __div_end
        push  {r0, r1}
        // variables
        mov   r0, #0        // quotient
        pop   {r1, r2}       // dividend / remainder, divisor
        mov   r3, #1        // bit field
 __div_shift:
        // shift divisor left until it exceeds dividend
        // the bit field will be shifted by one less
        cmp   r2, r1
        lslls r2, r2, #1
        lslls r3, r3, #1
        bls   __div_shift
__div_sub:
        // cmp sets the carry flag if r1 - r2 is positive, which is weird
        cmp   r1, r2
        // subtract divisor from the remainder if it was smaller
        // this also sets the carry flag since the result is positive
        subcs r1, r1, r2
        // add bit field to the quotient if the divisor was smaller
        addcs r0, r0, r3
        // shift bit field right, setting the carry flag if it underflows
        lsrs  r3, r3, #1
        // shift divisor right if bit field has not underflowed
        lsrcc r2, r2, #1
        // loop if bit field has not underflowed
        bcc   __div_sub
__div_end:
        sub   sp, fp, #4
        pop   {fp, pc}
```

Reference:
* [Simple division algorithm for ARM assembler](https://www.virag.si/2010/02/simple-division-algorithm-for-arm-assembler/)
* [Assembly Language - Division](http://www.tofla.iconbar.com/tofla/arm/arm02/index.htm)
* Pull request #56

jserv:
Video: [computers suck at division (a painful discovery)](https://youtu.be/ssDBqQ5f5_0).

PR #126, opened 2024-03-28
Generate a sequence of instructions for divisions on Arm targets

DrXiao:
Division instruction doesn't exist in ARMv7-A  (#46 ), but the current implementation of shecc still generates it to compute the result for division or modulo operations.

Thus, here adjusts arm-codegen so that shecc will generate an instruction sequence to achieve integer division without relying on any division instruction.


jserv (changes requested):
1. Add new command line option `+m`, which indicates the use of hardware integer multiplication and division instructions.
2. Extend `tests/driver.sh` for more integer division coverage.

DrXiao:
@jserv I understand that we can add a `+m` option to determine whether shecc generates hardware integer multiplication and division instructions.

However, I consider shecc can still generate multiplication instructions for ARM even if the `+m` option is closed. Because ARMv7-A inherently includes multiplication instruction, we can generate it directly without any concern.

Therefore, for the implementation of the `+m` option, I think that I can just handle whether the hardware integer division instructions are generated for ARM.

DrXiao:
The reference implementation mentioned in #46 assumes that the dividend and divisor are unsigned integers, but it may fail if at least one of them is a signed integer.

Can I suppose that shecc will not encounter any signed integer division and just handle unsigned integer division?
Or do we must let shecc handle both of them?

jserv:
By the way, I enjoy this article [Fast(est) Double-Word Integer Division](https://skanthak.hier-im-netz.de/integer.html).

DrXiao:
I have a new question about the **right shift operation**. During these days, I test the following code:
```c
/* test.c */
int main() {
    printf("%d\n", -2 >> 31);
    return 0;
}
```
In C99, It mentions that it is an **implementation-defined** behavior for using right shift operation on signed integers.

If we use the gcc and clang to compile and execute it, we can all obtain the output `-1`:
```
$ gcc -o test test.c
$ ./test
-1
$ clang -o test test.c
$ ./test
-1
```
I believe it may be a convention that many compilers use **arithmetic right shift** instructions for signed integers. 

But, when using shecc to test the right shift operation, we will get the output `1` because shecc uses **logical right shift** instructions for signed integers:
```
$ qemu-arm test
1
```

Should we instruct shecc to generate arithmetic right shift instructions as well?

jserv:
> In C99, It mentions that it is an **implementation-defined** behavior for using right shift operation on signed integers.
[...]
> Should we instruct shecc to generate arithmetic right shift instructions as well?

@vacantron, can you comment this?

jserv (changes requested):
Rebase the latest `master` branch and resolve the conflicts.

jserv (changes requested):
Rebase the latest `master` branch for bool specific fixes.

jserv:

DrXiao:
@jserv May I ask why this PR can be merged? Or will this PR be reopened and requested to be improved again in the future?

jserv:
> May I ask why this PR can be merged? Or will this PR be reopened and requested to be improved again in the future?

The latter.

Issue #123, opened 2024-03-25
Unable to self compile stage 1


ChAoSUnItY:
In commit 5f4f6d65b6670b8ac01bde5055da6075ed09a10a, by applying the following command `make distclean config ARCH=arm` and then `make` to build stage 1 compiler, the compiler will report unexpected token error at unreasonable position:
```
Error Unexpected token at source location 20771
    ref_block_t *head, *tail;
                       ^ Error occurs here
```

jserv:
It was my mistake to push the unintended changes. I will create the issues associated with the declaration changes.

PR #128, opened 2024-04-03
Fix _Bool type register allocation

### LINKED ISSUE #122: Support macros defined in <stdbool.h>

ChAoSUnItY:
- Resolves #122, fix `_Bool` type register allocation and utilize back all `bool` usages introduced in PR #123.

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -83,16 +83,16 @@ typedef enum {
 char token_str[MAX_TOKEN_LEN];
 token_t next_token;
 char next_char;
-int skip_newline = 1;
+bool skip_newline = true;
 
-int preproc_match;
+bool preproc_match;
 
 /* Point to the first character after where the macro has been called. It is
  * needed when returning from the macro body.
  */
 int macro_return_idx;
 
-int is_whitespace(char c)
+bool is_whitespace(char c)
 {
     return c == ' ' || c == '\t';
 }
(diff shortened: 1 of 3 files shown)


jserv (changes requested):
Specify the return type to `bool` for function `read_body_assignment` and `read_global_assignment`.

PR #131, opened 2024-06-13
Fix bad logical-and implementation

### LINKED ISSUE #120: Fail to self-host
The stage-1 and stage-2 ELF are not identical. Check them with the following `diff` command:

```shell
$ diff <(arm-linux-gnueabihf-objdump -d out/shecc-stage1.elf) <(arm-linux-gnueabihf-objdump -d out/shecc-stage2.elf)
```

output:

```diff
35856,35861c35856,35861
<    33074:     e59d5000        ldr     r5, [sp]
<    33078:     e58d50bc        str     r5, [sp, #188]  ; 0xbc
<    3307c:     e59d6000        ldr     r6, [sp]
<    33080:     e58d60c0        str     r6, [sp, #192]  ; 0xc0
<    33084:     e59d7000        ldr     r7, [sp]
<    33088:     e58d70c4        str     r7, [sp, #196]  ; 0xc4
---
>    33074:     e59d6000        ldr     r6, [sp]
>    33078:     e58d60bc        str     r6, [sp, #188]  ; 0xbc
>    3307c:     e59d7000        ldr     r7, [sp]
>    33080:     e58d70c0        str     r7, [sp, #192]  ; 0xc0
>    33084:     e59d2000        ldr     r2, [sp]
>    33088:     e58d20c4        str     r2, [sp, #196]  ; 0xc4
36201,36206c36201,36206
<    335d8:     e59d5000        ldr     r5, [sp]
<    335dc:     e58d50bc        str     r5, [sp, #188]  ; 0xbc
<    335e0:     e59d6000        ldr     r6, [sp]
<    335e4:     e58d60c0        str     r6, [sp, #192]  ; 0xc0
<    335e8:     e59d7000        ldr     r7, [sp]
<    335ec:     e58d70c4        str     r7, [sp, #196]  ; 0xc4
---
>    335d8:     e59d6000        ldr     r6, [sp]
>    335dc:     e58d60bc        str     r6, [sp, #188]  ; 0xbc
>    335e0:     e59d7000        ldr     r7, [sp]
>    335e4:     e58d70c0        str     r7, [sp, #192]  ; 0xc0
>    335e8:     e59d2000        ldr     r2, [sp]
>    335ec:     e58d20c4        str     r2, [sp, #196]  ; 0xc4
36463,36468c36463,36468
<    339f0:     e59d5000        ldr     r5, [sp]
<    339f4:     e58d50bc        str     r5, [sp, #188]  ; 0xbc
<    339f8:     e59d6000        ldr     r6, [sp]
<    339fc:     e58d60c0        str     r6, [sp, #192]  ; 0xc0
<    33a00:     e59d7000        ldr     r7, [sp]
<    33a04:     e58d70c4        str     r7, [sp, #196]  ; 0xc4
---
>    339f0:     e59d6000        ldr     r6, [sp]
>    339f4:     e58d60bc        str     r6, [sp, #188]  ; 0xbc
>    339f8:     e59d7000        ldr     r7, [sp]
>    339fc:     e58d70c0        str     r7, [sp, #192]  ; 0xc0
>    33a00:     e59d2000        ldr     r2, [sp]
>    33a04:     e58d20c4        str     r2, [sp, #196]  ; 0xc4
```

### LINKED ISSUE #137: Implement short-circuit evaluation of `&&` operator
According to the C99 standard (6.5.13):

> The `&&` operator shall yield 1 if both of its operands compare unequal to 0; otherwise, it yields 0. The result has type `int`.
> Unlike the bitwise binary `&` operator, the `&&` operator guarantees left-to-right evaluation; there is a sequence point after the evaluation of the first operand. If the first operand compares equal to 0, the second operand is not evaluated.

The latter operand should not be evaluated if the former is the expression of 0.

nosba0957:
The previous implementation of the logical-and operation produced incorrect result. This error also affected the funtionality of if statements using logical-and.

This implementation follows the C99 standard for logical-and, ensuring left-to-right evaluation. If the first operand is zero, the second operand is not evaluated.

Close #137


jserv:
Can you add some test cases in `tests/driver.sh` to reflect the proposed changes?

DrXiao:
I use the following code to test the logical-and operation: 

```c
/* test.c */
#include <stdio.h>

int func(int x)
{
    if (x >= 0) {
        printf("x >= 0\n");
    }   
    return x;
}   

int main()
{
    int ret = 0;
    
    ret = 1 && func(5);
    if (ret)                   
        printf("%d\n", ret);

    printf("============\n");
        
    ret = 0 && func(5);
    if (ret)
        printf("%d\n", ret);

    return 0;
}

```
If we use GCC or Clang to compile and execute the code, both yield the same output:
```
$ gcc -o test-gcc test.c && ./test-gcc
x >= 0
1
============
```
```
$ clang -o test-clang test.c && ./test-clang
x >= 0
1
============
```
In the `main` function, for the logical-and operation, because the first operand of the second `if` statement is zero, its second operand, which calls the function `func()`, will not be evaluated. (This follows the **left-to-right evaluation** rule.)

However, if we use shecc with the modified implementation to generate an executable with Arm32, the result is unexpected:
```
$ qemu-arm out/shecc-stage2.elf -o test-shecc-stage2 test.c && qemu-arm test-shecc-stage2
x >= 0
5
============
x >= 0
```
For the second `if` statement, although the first operand is zero, the second operand (`func(5)`) is still executed. Besides, **the result of the logical-and should be 0 or 1**, but the proposed changes may cause the second operand's value being returned.

Therefore, I think we can ensure the correctness of the logical-and behavior's result, but we still need to have more effort to implement the **left-to-right evaluation** rule for logical-and and logical-or operations.

nosba0957:
Sure, I'll add test cases to `tests/driver.sh`.

@DrXiao, Thank you for your feedback. I've decided to first ensure the correctness of the logical-and operation. Currently, left-to-right evaluation cannot be achieved. I will revise the commit messages.

jserv:
> Currently, left-to-right evaluation cannot be achieved.

Per the C99 standard, left-to-right evaluation should be guaranteed. You should investigate the root cause preventing the proposed changes from being compliant with the C standard.

src/arm-codegen.c
```diff
@@ -443,8 +445,14 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
         emit(__mov_i(__EQ, rd, 1));
         return;
     case OP_log_and:
-        /* FIXME: bad logical-and instruction */
-        emit(__and_r(__AL, rd, rn, rm));
+        /* TODO: Short-circuit evaluation */
```
vacantron, line 448: You do have the short circuit evaluation here. The previously mentioned issue is caused by parser, not codegen.
You can remove this TODO here.

src/riscv-codegen.c
```diff
@@ -340,8 +340,12 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
         emit(__xori(rd, rd, 1));
         return;
     case OP_log_and:
-        /* FIXME: bad logical-and instruction */
-        emit(__and(rd, rs1, rs2));
+        /* TODO: Short-circuit evaluation */
```
vacantron, line 343: Ditto.

nosba0957:
@vacantron, I'm currently modifying the parser to achieve left-to-right evaluation. But if I start modifying the implementation from the parser, the changes to the codegen might not be necessary.

vacantron:
> @vacantron, I'm currently modifying the parser to achieve left-to-right evaluation. But if I start modifying the implementation from the parser, the changes to the codegen might not be necessary.

That's okay. You can close this pull request since then.

jserv (changes requested):
Use `git rebase` instead of `git merge`.

nosba0957:
In `src/ssa.c`, within the `insert_phi_insn` function, why we didn't link the `prev` pointer between the original instruction and the new phi-func?
This can cause errors in `append_unwound_phi_insn`. For example, if a basic block has two instructions:
```
phi %a
branch %b
```
When we execute `append_unwound_phi_insn`, some phi-functions need to insert new unwind-phi instruction in this basic block. If a branch instruction is detected at the end of basic block, it will insert unwind-phi instruction before the branch. However, we do not link `prev` from `branch %b` to `phi %a`, so `phi %a` will be directly replaced when inserting the unwind-phi instruction.
When attempting to link the `prev` pointer in `insert_phi_insn`, it ends with segmentation fault. Therefore,  I attempted to modify `append_unwound_phi_insn` to search for a place for unwind_phi instruction from head of `insn_list`. It finally runs correctly.

This problem can be reproduced by removing the modification in `src/ssa.c` and executing a simple example of do-while loop with logical-and condition expression.

vacantron:
> In `src/ssa.c`, within the `insert_phi_insn` function, why we didn't link the `prev` pointer between the original instruction and the new phi-func?

The property `prev` was added later and I forgot to update the related code. Thank you to point it out.

vacantron (changes requested):
There is a snapshot test under the `tests/` directory to watch the consistence of the work of lever/parser. Since they have been changed, the corresponding snapshots need to be updated together.

You can run `tests/update-snapshots.sh` under the ARM config.

nosba0957:
> The property `prev` was added later and I forgot to update the related code. Thank you to point it out

So should I keep the modifications in `append_unwound_phi_insn` for now?

> You can run tests/update-snapshots.sh under the ARM config.

Thanks for your friendly reminder.

vacantron:
> So should I keep the modifications in append_unwound_phi_insn for now?

You can add a comment like `FIXME: ...` to indicate the workaround.

src/parser.c
```diff
@@ -973,20 +976,37 @@ void read_expr_operand(block_t *parent, basic_block_t **bb)
...
+    strcpy(land_else_label, gen_label());
+
     read_expr_operand(parent, bb);
 
     opcode_t op = get_operator();
     if (op == OP_generic || op == OP_ternary)
         return;
-
-    oper_stack[oper_stack_idx++] = op;
+    else if (op == OP_log_and) {
```
jserv, line 1004: Replace "else if" with "if" statement.

src/ssa.c
```diff
@@ -625,14 +626,25 @@ void append_unwound_phi_insn(basic_block_t *bb, var_t *dest, var_t *rs)
...
         if (tail->opcode == OP_branch) {
-            if (tail->prev) {
+            if (head->opcode == OP_phi) {
+                /* FIXME: This temporary implementation handles the case that if
+                 * tail->prev is OP_phi. The error in insert_phi_insn() is due
+                 * to not linking the `prev` pointer of the next instruction to
+                 * the newly added phi instruction. Once this error is fixed, we
+                 * can remove the first if statement.
+                 */
+                while (head->next->opcode != OP_branch)
```
jserv, line 644: Can you avoid traversing the whole elements since we just want to figure out the branch operation at first glance?

nosba0957: Should I still modify this part of the code, or should I keep the current commit as is and wait for @vacantron 's fix?

vacantron: You can cherry-pick the commit at https://github.com/vacantron/shecc/commit/5fd2fbe211cc9bf49a84fb768169b38ea841859c and update your change.

vacantron:
@nosba0957,I have opened the #137. You can redirect this pull request to that issue.

jserv (changes requested):

vacantron (approved):

jserv:
Thank @nosba0957 for contributing!

PR #132, opened 2024-06-16
Fix some typos

DrXiao:
After reviewing PR #126, I found that I wrote some typos and left unnecessary brackets when using the `if` statements, so this commit fixes them in `arm-codegen.c` and `arm.c`.


jserv (changes requested):
This pull request should fix typo without style changes.

PR #133, opened 2024-06-18
Implement multiplication for RV32I

DrXiao:
Previously, shecc would just generate hardware multiplication instructions for multiplication operations.

This commit adjusts `riscv-codegen.c` so that shecc can generate a sequence of instructions to achieve multiplication without any RV32M instruction if the '+m' option is disable.

Related: #46

### DIFF
--- a/tests/driver.sh
+++ b/tests/driver.sh
@@ -592,6 +592,39 @@ int main()
 }
 EOF
 
+# Multiplication for signed integers
+try_output 0 "35 -35 -35 35" << EOF
+int main()
+{
+    printf("%d %d %d %d\n", 5 * 7, 5 * (-7), (-5) * 7, (-5) * (-7));
+    return 0;
+}
+EOF
+
+try_output 0 "-212121 -535050 336105 666666666" << EOF
+int main()
+{
+    printf("%d %d %d %d\n", (-333) * 637, 1450 * (-369), 37345 * 9, (-111111111) * (-6));
+    return 0;
+}
+EOF
+
+try_output 0 "1073676289 -131071 30" << EOF
+int main()
+{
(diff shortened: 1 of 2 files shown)


tests/driver.sh
```diff
@@ -592,6 +592,39 @@ int main()
 }
 EOF
 
+# Multiplication for signed integers
+try_output 0 "35 -35 -35 35" << EOF
+int main()
+{
+    printf("%d %d %d %d\n", 5 * 7, 5 * -7, -5 * 7, -5 * -7);
```
jserv, line 599: Always use `(-5)` and `(-7)`.

DrXiao: Does it mean that I should enclose each negative integer in brackets?
```shell
# Multiplication for signed integers
try_output 0 "35 -35 -35 35" << EOF
int main()
{
    printf("%d %d %d %d\n", 5 * 7, 5 * (-7), (-5) * 7, (-5) * (-7));
    return 0;
}
EOF
```
Regardless, I will continue to work it tomorrow.

jserv: > Does it mean that I should enclose each negative integer in brackets?

Yes, it is more friendly.

jserv (changes requested):
Append "Close #46" at the end of git commit messages.

DrXiao:
This pull request just resolves implementing multiplication for RV32I, but division is still dependent on RV32M.
I think this issue should be closed by the next pull request implementing division for RV32I, but not closed by this pull request.

jserv (changes requested):
Improve the git commit messages by fix grammar mistakes.

DrXiao:
I apologize for using improper git commit messages. I will refine my message and update it.

jserv:

PR #134, opened 2024-06-20
Implement division and modulo for RV32I

### LINKED ISSUE #46: Support Arm/RISC-V targets without integer division instructions

DrXiao:
This commit allows shecc to generate a sequence of instructions to perform division or modulo for RISC-V targets with RV32I only.

Close #46

### DIFF
--- a/src/riscv-codegen.c
+++ b/src/riscv-codegen.c
@@ -50,8 +50,6 @@ void update_elf_offset(ph2_ir_t *ph2_ir)
     case OP_indirect:
     case OP_add:
     case OP_sub:
-    case OP_div:
-    case OP_mod:
     case OP_lshift:
     case OP_rshift:
     case OP_gt:
@@ -69,6 +67,13 @@ void update_elf_offset(ph2_ir_t *ph2_ir)
             elf_offset += 4;
         else
             elf_offset += 52;
+        return;
+    case OP_div:
+    case OP_mod:
+        if (hard_mul_div)
+            elf_offset += 4;
+        else
+            elf_offset += 104;
         return;
     case OP_load_data_address:
     case OP_neq:
@@ -151,6 +156,11 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
     int rs1 = ph2_ir->src0 + 10;
     int rs2 = ph2_ir->src1 + 10;
     int ofs;
+
+    /* Prepare the variables to reuse the same code for
+     * the instruction sequence of division and modulo.
+     */
+    rv_reg soft_div_rd = __t0, divisor_mask = __t1;
 
     switch (ph2_ir->op) {
     case OP_define:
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
The code generator for division and modulo has only 2 instructions in difference.
```diff
--- codegen-div.c
+++ codegen-mod.c
@@ -7,7 +7,7 @@
             emit(__srai(__t1, __t3, 31));
             emit(__add(__t3, __t3, __t1));
             emit(__xor(__t3, __t3, __t1));
-            emit(__xor(__t5, __t0, __t1));
+            emit(__addi(__t5, __t0, 0));
             /* Unsigned integer division */
             emit(__addi(__t0, __zero, 0));
             emit(__addi(__t1, __zero, 1));
@@ -23,7 +23,7 @@
             emit(__srli(__t1, __t1, 1));
             emit(__srli(__t3, __t3, 1));
             emit(__bne(__t1, __zero, -20));
-            emit(__addi(rd, __t0, 0));
-            /* Handle the correct sign for the quotient */
+            emit(__addi(rd, __t2, 0));
+            /* Handle the correct sign for the remainder */
             emit(__beq(__t5, __zero, 8));
             emit(__sub(rd, __zero, rd));
```

Thus, we can introduce a helper function to generate div/mod emulation instructions, enhancing code reusability.

jserv:
> Thus, we can introduce a helper function to generate div/mod emulation instructions, enhancing code reusability.

A similar case for div/mod emulation instruction generation exists in the Arm backend. Let's refine the RISC-V backend in this pull request and address the Arm backend in another.

DrXiao:
Yes, the proposed changes can be improved to enhance the code reusability. I will refine the RISC-V backend first and submit another pull request for refining the Arm backend.

DrXiao:
Rather than adding a helper function, I thought of another approach that adds two local variables in `emit_ph2_ir()` and uses `if` statements to achieve reusability.

src/riscv-codegen.c
```diff
@@ -307,10 +317,51 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
         }
         return;
     case OP_div:
-        emit(__div(rd, rs1, rs2));
-        return;
     case OP_mod:
-        emit(__mod(rd, rs1, rs2));
+        if (hard_mul_div)
```
jserv, line 321: Rewrite as following:
```c
        if (hard_mul_div) {
            if (ph2_ir->op == OP_div)
                emit(__div(rd, rs1, rs2));
            else
                emit(__mod(rd, rs1, rs2));
            return;
        }
        /* div/mod emulation */
        if (ph2_ir->op == OP_mod) {
            ...
```

src/riscv-codegen.c
```diff
@@ -152,6 +157,11 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
     int rs2 = ph2_ir->src1 + 10;
     int ofs;
 
+    /* Prepare the variables to reuse the same code for
+     * the instruction sequence of division and modulo.
+     */
+    rv_reg soft_div_rd = __t0, soft_div_rs2_sign_reg = __t1;
```
jserv, line 163: `soft_div_rs2_sign_reg` looks weird. Can you shorten it?

src/riscv-codegen.c
```diff
@@ -307,10 +317,52 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
...
+        if (hard_mul_div) {
+            if (ph2_ir->op == OP_div)
+                emit(__div(rd, rs1, rs2));
+            else
+                emit(__mod(rd, rs1, rs2));
+            return;
+        }
+        /* div/mod emulation */
+        if (ph2_ir->op == OP_mod) {
+            /* If the operation is modulo, the result will be stored
```
jserv, line 330: Proofread as following:
```c
            /* If the requested operation is modulo, the result will be stored
             * in __t2. The sign of the divisor is irrelevant for determining
             * the result's sign.
             */
```

src/riscv-codegen.c
DrXiao: How about renaming it as  `divisor_mask_reg`? 

At the beginning of the emulation, it obtains the absolute values of the dividend and divisor using the following approach.
```c
int32_t abs(int32_t x) 
{
    int32_t mask = x >> 31;       /* arithmetic right shift */
    return (x + mask) ^ mask
}
```
Therefore, the register  `__t1` will store `0` or `0xFFFFFFFF` initially, indicating that it serves as the mask for the divisor. 
(Similarly, `__t0` will be the mask for the dividend.)

If the requested operation is division, `divisor_mask_reg` will be `__t1` and helps to determine the result's sign. Otherwise,  for the modulo operation, because it only needs the dividend's mask to determine the result's sign, `divisor_mask_reg` becomes unnecessary and is assigned to `__zero` so that it doesn't affect the result's sign.

src/riscv-codegen.c
DrXiao: I will rewrite it later.

src/riscv-codegen.c
jserv: Simply name it `divisor_mask` as it is intended to behave as its name suggests.

jserv (changes requested):
Execute `git rebase -i` to squash the commits.

jserv:

PR #136, opened 2024-06-30
Improve the division emulation in the Arm backend

### LINKED ISSUE #135: Enhance the implementation of division emulation in the Arm backend
Currently, the division emulation in the Arm backend has two potential issues to be resolved:

1. Enhance code reusability:
    There are only 2 instructions different for div/mod emulation, and it can be refined to reuse the same code.
    ```diff
        /* Obtain absolute values of dividend and divisor */
        emit(__srl_amt(__AL, 0, arith_rs, __r8, rn, 31));
        emit(__add_r(__AL, rn, rn, __r8));
        emit(__eor_r(__AL, rn, rn, __r8));
        emit(__srl_amt(__AL, 0, arith_rs, __r9, rm, 31));
        emit(__add_r(__AL, rm, rm, __r9));
        emit(__eor_r(__AL, rm, rm, __r9));
    -   emit(__eor_r(__AL, __r10, __r8, __r9));
    +   emit(__mov_r(__AL, __r10, __r8));
        /* ... */
    -   emit(__mov_r(__AL, rd, __r9));
    -   /* Handle the correct sign for quotient */
    +   emit(__mov_r(__AL, rd, rn));
    +   /* Handle the correct sign for remainder */
        emit(__cmp_i(__AL, __r10, 0));
        emit(__rsb_i(__NE, rd, 0, rd));
    ```
2. There has another potential issue for the current implementation: **the dividend and divisor will lose after the emulation.**
     
    If observing the implementation, `rn` and `rm`, which storing the dividend and divisor, will be shifted to complete the emulation, and the original value will lost at the end. But, for example, an executable compiled by shecc may has the following instruction sequence:
    * with `+m` option (use hardware div/mod instruction)
    ```
    .....
    .....
    div r2, r0, r1 
    ldr r3, r0           ; Load a word by r0 (the dividend).
    .....
    ```
    * without `+m` option (use div/mod emulation)
    ```
    
    ....
    asr     r8, r0, #31
    add     r8, r8, r0
    eor     r8, r8, r0
    .....
    .....
    cmp     sl #0
    rsbne   r2, r2, #0
    ldr     r3, r0        ; Load a word by r0 (the dividend). 
    ......                ; But r0 becomes no meaning because the emulation.
    ......
    ```
    After a division, the subsequent insturction(s) may use the register storing the dividend (or divisor) to perform the operation. With hardware instruction support, it will have no any problem. However, it may be failed for the subsequent instruction(s) because `rn` and `rm` become no meaning after the emulation.

    Therefore, the implementation should be improved to restore the dividend and divisor after the emulation.

DrXiao:
This commit improves the division emulation in the Arm backend to enhance code reusability and eliminate potential issues.

Close #135

### DIFF
--- a/src/arm.c
+++ b/src/arm.c
@@ -33,11 +33,13 @@ typedef enum {
     arm_sub = 2,
     arm_rsb = 3,
     arm_add = 4,
+    arm_ldm = 9,
     arm_teq = 9,
     arm_cmp = 10,
     arm_orr = 12,
     arm_mov = 13,
-    arm_mvn = 15
+    arm_mvn = 15,
+    arm_stmdb = 16
 } arm_op_t;
 
 /* Condition code
@@ -281,6 +283,16 @@ int __sw(arm_cond_t cond, arm_reg rd, arm_reg rn, int ofs)
 int __sb(arm_cond_t cond, arm_reg rd, arm_reg rn, int ofs)
 {
     return arm_transfer(cond, 0, 1, rn, rd, ofs);
+}
+
+int __stmdb(arm_cond_t cond, int w, arm_reg rn, int reg_list)
+{
+    return arm_encode(cond, arm_stmdb + (0x2 << 6) + (w << 1), rn, 0, reg_list);
+}
+
+int __ldm(arm_cond_t cond, int w, arm_reg rn, int reg_list)
+{
+    return arm_encode(cond, arm_ldm + (0x2 << 6) + (w << 1), rn, 0, reg_list);
 }
 
 int __b(arm_cond_t cond, int ofs)
--- a/src/arm-codegen.c
+++ b/src/arm-codegen.c
@@ -85,18 +85,16 @@ void update_elf_offset(ph2_ir_t *ph2_ir)
         elf_offset += 4;
         return;
     case OP_div:
-        if (hard_mul_div) {
-            elf_offset += 4;
-        } else {
-            elf_offset += 104;
-        }
-        return;
     case OP_mod:
         if (hard_mul_div) {
-            elf_offset += 12;
-        } else {
-            elf_offset += 104;
+            if (ph2_ir->op == OP_div)
+                elf_offset += 4;
+            else
+                elf_offset += 12;
+            return;
         }
+        /* div/mod emulation's offset */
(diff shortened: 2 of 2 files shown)


src/arm.c
```diff
@@ -283,6 +285,16 @@ int __sb(arm_cond_t cond, arm_reg rd, arm_reg rn, int ofs)
     return arm_transfer(cond, 0, 1, rn, rd, ofs);
 }
 
+int __push(arm_cond_t cond, int reg_list)
```
jserv, line 288: Both `push` and `pop` are pseudo-instructions. Can you avoid them?

DrXiao: After reading [ARM Architecture Reference Manual](https://developer.arm.com/documentation/ddi0406/latest/), I understand that `push` and `pop` are equivalent to `stmdb` and `ldmia`, respectively. But I don't see any description stating that `push` and `pop` are pseudo-instructions.

May I ask where it describes `push` and `pop` as pseudo-instructions? Although I can replace them with `stmdb` and `ldmia`, I want to understand `push` and `pop` further first before improving the implementation.

jserv: > May I ask where it describes `push` and `pop` as pseudo-instructions? 

See https://developer.arm.com/documentation/dui0489/i/arm-and-thumb-instructions/pop

Both `__push` and `__pop` are too generic, likely leading to unexpected symbol conflicts.

DrXiao: >Both push and pop are pseudo-instructions. Can you avoid them?

I think I misunderstood at the beginning. Originally,  I thought it was related to code generation for Arm32 output.

However, the requested changes just entail that I should replace `__push` and `__pop` with `__stmdb` and `__ldm`  in `arm.c` and `arm-codegen.c`. The purpose is to avoid symbol conflicts during compilation.

Is my understanding correct?

jserv: > However, the requested changes just entail that I should replace `__push` and `__pop` with `__stmdb` and `__ldm` in `arm.c` and `arm-codegen.c`. The purpose is to avoid symbol conflicts during compilation.

Exactly. Unless we can change symbol visibility on demand, we must handle symbol conflicts with care.

jserv:

PR #140, opened 2024-07-15
Speed up malloc/free routines

### LINKED ISSUE #139: Significantly increased bootstrap time in recent build
@vacantron 's analysis indicates that the previous bootstrap took 10 seconds; with the introduction of #131, it increases to 40 seconds. However, if the `free()` function's body is emptied and its corresponding actions are omitted, the time reduces to 3.5 seconds.

DrXiao:
This commit improves `malloc()` and `free()` in `lib/c.c` to eliminate a useless member from a structure and reduce execution time.

Refine the definition of `chunk_t` and improve the two functions using techniques similar to glibc.

For example, in glibc, it can obtain the starting address of a chunk by a macro called `mem2chunk()`, and the macro moves the given pointer backward directly to get the address of a chunk. Thus, the proposed changes use a similar approach to speed up `free()`.

Close #139


jserv (changes requested):
Avoid using [backticks](https://en.wikipedia.org/wiki/Backtick) in git commit messages and comments to ensure compatibility with terminal emulators that may not render the backtick character properly.

Instead, use the pair `'` or `"` (quotation marks).

lib/c.c
```diff
@@ -529,11 +529,16 @@ int fputc(int c, FILE *stream)
 /* Non-portable: Assume page size is 4KiB */
 #define PAGESIZE 4096
 
+#define CHUNK_SIZE_FREED_BIT_MASK 1
```
jserv, line 532: Simply name the identifiers as mask, which represents bit mask. That is, `CHUNK_SIZE_FREED_MASK`.

lib/c.c
```diff
@@ -529,11 +529,16 @@ int fputc(int c, FILE *stream)
...
+#define CHUNK_SIZE_SZ_BITS_MASK 0xFFFFFFFE
+#define CHUNK_GET_SIZE(size) (size & CHUNK_SIZE_SZ_BITS_MASK)
+#define CHUNK_GET_FREED_FLAG(size) (size & CHUNK_SIZE_FREED_BIT_MASK)
```
jserv, line 535: Rename it to `IS_CHUNK_GET_FREED` or something alike.

lib/c.c
```diff
@@ -529,11 +529,16 @@ int fputc(int c, FILE *stream)
 /* Non-portable: Assume page size is 4KiB */
 #define PAGESIZE 4096
 
+#define CHUNK_SIZE_FREED_MASK 1
+#define CHUNK_SIZE_SZ_MASK 0xFFFFFFFE
+#define CHUNK_GET_SIZE(size) (size & CHUNK_SIZE_SZ_MASK)
+#define IS_CHUNK_GET_FREED(size) (size & CHUNK_SIZE_FREED_MASK)
+#define CHUNK_SET_FREED(size) (size | CHUNK_SIZE_FREED_MASK)
```
jserv, line 536: How about rewriting `CHUNK_SET_FREED` as following?
```c
#define CHUNK_SET_FREED(size) \
    size |= CHUNK_SIZE_FREED_MASK
```

jserv:
Improve the ASCII art as following:
```
                     |<--  usable memory  -->|
+------+------+------+-------- ... ----------+
| next | prev | size |          :            |
+------+------+------+-------- ... ----------+
^                    ^
|                    |
chunk              (chunk + 1)
                    the starting address returned by malloc().
```

jserv (changes requested):
Provide some timing measurements for the proposed changes.

lib/c.c
DrXiao: I encounter some trouble when modifying the macro.

Then, I use the following code to test a function-like macro:
```c
/* test.c */
#define MACRO(a) a += 10

int main()
{
    int x = 10;
    printf("%d\n", x);
    MACRO(x);
    printf("%d\n", x);

    return 0;
}
```
```
Error Unrecognized statement token at source location 16802
#define MACRO(a) a += 10
                   ^ Error occurs here
Abnormal program termination
```
This implies that shecc cannot parse the macro containing compound operator(s) correctly.

jserv: The current parser lacks of the ability to deal with compound assignment operators. So, you have to change the statements.

DrXiao: Because the current parser also lacks of the ability to handle assignment operator. Therefore, I think I will use two functions called `chunk_set_freed()` and `chunk_clear_freed()` to perform the operations instead of using the original macros.

```c
void chunk_set_freed(chunk_t *chunk) 
{
        chunk->size |= CHUNK_SIZE_FREED_MASK;
}

void chunk_clear_freed(chunk_t *chunk) 
{
        chunk->size &= CHUNK_SIZE_SZ_MASK;
}
```

jserv: > I think I will use two functions called `chunk_set_freed()` and `chunk_clear_freed()` to perform the operations instead of using the original macros.

It looks good to me.

DrXiao:
I will continue to improve my commit messages and changes later, but I provide my measurements here first.

Here is the information on my CPU.
```
$ lscpu
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         39 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  4
  On-line CPU(s) list:   0-3
Vendor ID:               GenuineIntel
  Model name:            Intel(R) Core(TM) i5-7300HQ CPU @ 2.50GHz
    CPU family:          6
    Model:               158
    Thread(s) per core:  1
    Core(s) per socket:  4
    Socket(s):           1
    Stepping:            9
    CPU max MHz:         3500.0000
    CPU min MHz:         800.0000
...
```

My measurements use `time` command to build shecc for Arm and RISC-V outputs and obtain their execution time.
```
$ make distclean && time make
$ make distclean && time make ARCH=riscv
```

On the `master` branch, it took about 26 seconds and 39 seconds to build the executables.
```
$ make distclean && time make
...
  SHECC out/shecc-stage2.elf

real    0m26.886s
user   0m24.716s
sys     0m2.096s

$ make distclean && time make ARCH=riscv
...
  SHECC out/shecc-stage2.elf

real    0m38.896s
user    0m36.528s
sys     0m2.332s
```

Then, after introducing the proposed changes, it was reduced to about 6.6 seconds and 18 seconds, respectively.
```
$ make distclean && time make
...
  SHECC out/shecc-stage2.elf

real    0m6.618s
user    0m4.283s
sys     0m2.178s

$ make distclean && time make ARCH=riscv
...
  SHECC out/shecc-stage2.elf

real    0m17.923s
user    0m15.464s
sys     0m2.254s
```

|                          |   master   |   fix-build (proposed changes)   |  Percentage of reduced time  |
|---------------------|---------------|-----------------------------------------------| ----------------|
| shecc (arm)    |  26.886 s  |    6.618 s                                         |  75.38%      |
| shecc (riscv)   |  38.896 s  |    17.923 s                                       |  53.92%      |

By the proposed changes, the execution time will be significantly reduced.

lib/c.c
jserv: > the current parser also lacks of the ability to handle assignment operator.

@ChAoSUnItY, The compound assignment operators consist of a binary operator and the simple assignment operator. At present, there is no support for compound assignment operators. Can you evaluate the feasibility to consolidate the parser for compound assignment operators?

jserv:

DrXiao:
By the way, because I use `chunk_set_freed()` and `chunk_clear_freed()`, which replace `CHUNK_SET_FREED()` and `CHUNK_CLEAR_FREED()` to set or clear the bit, the execution time may increase a little due to the overhead of function calls.

Thus, here is the newest result of measurements:
|                         |  master  |   proposed changes  |  Percentage of reduced time |
|--------------------|--------------|-------------------------------|-------------------------------------------|
| shecc (arm)   | 26.845 s |  6.435 s                        |  76.02%                                      |
| shecc (riscv)  | 38.819 s |  18.198s                       | 53.12%                                       |

However, the bootstrap time is still significantly reduced.

lib/c.c
DrXiao: Here are the examples that shecc cannot deal with.
```c
#define MACRO1(variable, val) \
               variable = variable + val + 10
#define MACRO2(variable, val) \
               variable += val + 10
```
These macros contain an assignment operator and a compound assignment operator, and shecc cannot parse them correctly.

jserv: > Here are the examples that shecc cannot deal with.

Create an issue for tracking.

ChAoSUnItY: > The compound assignment operators consist of a binary operator and the simple assignment operator. At present, there is no support for compound assignment operators. 

The current situation is that `read_body_assignment` is failed to realize the macro variable and thus returns false since it's possibly not a local variable nor a global variable, which ultimately results in the parser failed to parse current variable and terminated at the operator.
 
> Can you evaluate the feasibility to consolidate the parser for compound assignment operators?

Currently, it seems rather tricky to implement due to the technical debt of macro parsing design (we're not able to determine which syntax parsing we should use to parse macro parameter). Moreover, if we predicts it's a lvalue (which in this case it should be) and we use `read_lvalue` function to parse the parameter, the function requires a non-null `var_t`, which is impossible because we haven't even parse macro parameter's actual expression; if we predicts it's an expression, we'll lose the target address of the lvalue.

This probably could be resolved after we implement [tangle](https://github.com/ChAoSUnItY/tangle) back to shecc since macro expansion would only happens in lexer and parser doesn't have to predict what parsing strategy should be used for the expanded macro parameter.

PR #143, opened 2024-07-21
Fix calloc to return NULL on malloc failure

visitorckw:
Ensure calloc returns NULL when malloc fails to allocate memory. This change prevents potential undefined behavior when memory allocation fails.


jserv (changes requested):
Run `tests/update-snapshots.sh`

jserv:

PR #144, opened 2024-08-03
Fix Arm outputs for right shift operations

DrXiao:
As [the discussion](https://github.com/sysprog21/shecc/pull/126#issuecomment-2028550996) in the pull request (#126) mentioned, shecc generates logical right shift instructions for right shift operations on Arm targets currently. However, the most famous compilers such as GCC or Clang generate arithmetic right shift instructions for the operations.

Therefore, this pull request modifies the Arm backend so that the behavior of right shift operations is consistent with GCC and Clang.


vacantron (approved):

jserv:

Issue #88, opened 2023-12-07
Implement basic optimizations


jserv:
With the inclusion of SSA in the middle-end, it is now time to implement some common optimizations on the new SSA-based IR. These include constant folding, copy propagation, and dead code elimination.

jserv:
After @vacantron  implements the constant folding optimization, this task will be considered complete, signifying that shecc achieves the goal of implementing a minimal optimizing C compiler capable of compiling itself with proper optimizations.

vacantron:
Should we abjust the directory structure that making `src/ssa.c` focus on building and create another `src/optimizer.c` to include the existent `src/peephole.c` and the other optimizers ?

jserv:
> Should we abjust the directory structure that making `src/ssa.c` focus on building the SSA and create another `src/optimizer.c` to include the existent `src/peephole.c` and the other optimizers ?

Up to you. I expect the availability of basic optimizers existing with minimal efforts.

jserv:
So far, we have implemented a few optimizations:
* Common subexpression elimination
* Constant optimization
* Peephole optimization
* Dead code elimination

With these optimizations complete, we consider this task concluded. Dedicated issues will be created for any additional optimizations as they are planned.

PR #145, opened 2024-08-05
Implement dead code elimination

nosba0957:
The creation of RDF is based on reverse CFG. The method of building RDF is similar to building DF, but instead of iterating every `prev` nodes, it visit `next`, `then_` and `else_` nodes. The code appears to extend the for-loop to handle three cases (compared to the method of building DF). 
Due to the addition of RDF, several elements were added to the `struct basic_block`. However, `MAX_FILED` macro is not large enough, so it has been adjusted to 64.
DCE is implemented using mark-sweep algorithm. `dce_insn` marks the useful instruction and `dce_sweep` removes the operations that are not marked "useful". The `dce_insn` must execute after `cse` and `const_folding` are done because `dce_insn` may mark some instructions that are not handled by `cse` and `const_folding`. These instruction might be mistakenly mark "useful".

Related issue: #88

### DIFF
--- a/src/defs.h
+++ b/src/defs.h
@@ -15,7 +15,7 @@
 #define MAX_TYPE_LEN 32
 #define MAX_PARAMS 8
 #define MAX_LOCALS 1500
-#define MAX_FIELDS 32
+#define MAX_FIELDS 64
 #define MAX_FUNCS 512
 #define MAX_FUNC_TRIES 2160
 #define MAX_BLOCKS 2048
@@ -23,6 +23,7 @@
 #define MAX_IR_INSTR 40000
 #define MAX_BB_PRED 128
 #define MAX_BB_DOM_SUCC 64
+#define MAX_BB_RDOM_SUCC 256
 #define MAX_GLOBAL_IR 256
 #define MAX_LABEL 4096
 #define MAX_SOURCE 327680
@@ -173,6 +174,7 @@ struct var {
     int subscripts_idx;
     rename_t rename;
     ref_block_list_t ref_block_list; /* blocks which kill variable */
+    struct insn *last_assign;
     int consumed;
     bool is_ternary_ret;
     bool is_log_and_ret;
@@ -308,6 +310,8 @@ struct insn {
     var_t *rs1;
     var_t *rs2;
     int sz;
+    bool useful; /* Used in DCE process. Set true if instruction is useful. */
+    basic_block_t *belong_to;
     phi_operand_t *phi_ops;
     char str[64];
 };
@@ -352,6 +356,7 @@ struct basic_block {
     struct basic_block *then_; /* conditional BB */
     struct basic_block *else_;
     struct basic_block *idom;
+    struct basic_block *r_idom;
     struct basic_block *rpo_next;
     struct basic_block *rpo_r_next;
     var_t *live_gen[MAX_ANALYSIS_STACK_SIZE];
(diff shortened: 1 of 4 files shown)


src/defs.h
```diff
@@ -308,6 +310,8 @@ struct insn {
     var_t *rs1;
     var_t *rs2;
     int sz;
+    bool useful;
```
jserv, line 313: It would be better to define "useful."

src/ssa.c
```diff
@@ -261,6 +261,145 @@ void build_df()
...
+    }
+
+    pred->rdom_next[i++] = succ;
+    succ->rdom_prev = pred;
+    return true;
+}
+
+void bb_build_rdom(fn_t *fn, basic_block_t *bb)
+{
+    basic_block_t *curr = bb;
```
jserv, line 339: Rewrite with shorter `for` statement. That is,
```c
    for (basic_block_t *curr = bb; curr != fn->exit; curr = curr->r_idom) {
        if (!rdom_connect(curr->r_idom, curr))
            break;
    }
```

src/ssa.c
```diff
@@ -1170,14 +1310,170 @@ bool const_folding(insn_t *insn)
...
+        }
+    }
+}
+
+void dce_sweep()
+{
+    for (fn_t *fn = FUNC_LIST.head; fn; fn = fn->next) {
+        for (basic_block_t *bb = fn->bbs; bb; bb = bb->rpo_next) {
+            for (insn_t *insn = bb->insn_list.head; insn; insn = insn->next) {
+                if (!insn->useful) {
```
jserv, line 1424: Reduce the indentions. That is, rewrite as following:
```c
    if (insn->useful)
        continue;
    ...
```

src/defs.h
nosba0957: Do you mean to add comments to this section of the code to make it clearer?

jserv: > Do you mean to add comments to this section of the code to make it clearer?

Exactly. Otherwise, it is hard to figure out what "useful" is.

src/ssa.c
```diff
@@ -261,6 +261,143 @@ void build_df()
...
+    }
+    free(args);
+}
+
+void bb_build_rdf(fn_t *fn, basic_block_t *bb)
+{
+    UNUSED(fn);
+
+    int cnt = 0;
+    cnt = cnt + (bb->next ? 1 : 0);
```
jserv, line 364: Rewrite as following:
```c
if (bb->next)
    cnt++;
if (bb->then_)
    cnt++
if (bb->else_)
    cnt++;
```

src/defs.h
```diff
@@ -308,6 +310,8 @@ struct insn {
     var_t *rs1;
     var_t *rs2;
     int sz;
+    bool useful; /* Used in DCE process. Set true if instruction is useful. */
+    basic_block_t *belong_to_bb;
```
jserv, line 314: How about naming it `belong_to`?

jserv:
I defer to @vacantron for confirmation.

src/defs.h
nosba0957: No problem, I will make the change. I originally wanted to distinguish it from the `belong_to` within the `basic_block_t`.

jserv (changes requested):
Remove incorrect description "51,357 lines instructions." You should explicitly mention 51,357 instructions" for certain ISA.

jserv:
Once the DCE is known to work properly, we can close #88 to declare the capability of performing basic optimizations.

vacantron (approved):

PR #146, opened 2024-08-05
Use branchless way to get rv_hi/rv_lo result

fennecJ:
We can avoid using conditional branch via bitwise operation.    
The value of `((val & (1 << 11)) << 1)` is exactly 4096 if `(val & (1 << 11))` is true.


jserv:

PR #147, opened 2024-08-09
Optimize __str_base10()

visitorckw:
The base-10 conversion method has been optimized by replacing the previous loop-based approach with a more efficient technique using bitwise operations. This new method simulates division and modulus operations by 10, avoiding multiplication, division, or modulus instructions, and results in notable performance gains. Experimental results show a considerable reduction in execution time across various ranges, with the new method consistently outperforming the old one. Additionally, the handling of negative numbers has been simplified.

| Range           | Old    | New    |
|-----------------|--------|--------|
| [0, 10)         | 0.473s | 0.293s |
| [0, 100)        | 0.619s | 0.434s |
| [0, 1000)       | 0.818s | 0.646s |
| [0, 10000)      | 1.715s | 0.902s |
| [0, 100000)     | 2.166s | 1.169s |
| [0, 1000000)    | 2.239s | 1.453s |
| [0, 10000000)   | 2.359s | 1.773s |
| [0, 100000000)  | 2.463s | 2.122s |


DrXiao (changes requested):
Please describe how you measure the execution time for your proposed changes and obtain these results.

> | Range |Old |New |
> | ----------|-----------|-------|
> | [0, 10) | 0.473s | 0.293s|
>  |... | ... | ... |
> | [0, 100000000) | 2.463s | 2.122s|

visitorckw:
I have updated the commit message to make the process of how I tested and obtained the experimental data clearer. For each test range, values within the range were used as inputs, and __str_base10() was called 100,000,000 times with these inputs uniformly distributed. The execution time was measured using the time command. The test code is as follows:

```
int x = 0, i;
const int range = 1000;
char s[INT_BUF_LEN];

for (i = 0; i < 100000000; i++) {
    __str_base10(s, x);
    x = (x + 1) % range;
}
```

DrXiao:
@visitorckw , I apologize that I forgot to read your commit messages first. But I have understood that the timing data is collected by your test code and `time` command. 
Thank you for your clarification!

lib/c.c
```diff
@@ -127,45 +129,20 @@ void __str_base10(char *pb, int val)
...
-        pb[INT_BUF_LEN - 3]++;
-    }
-    while (val >= 10) {
-        val -= 10;
-        pb[INT_BUF_LEN - 2]++;
-    }
-    while (val >= 1) {
-        val -= 1;
-        pb[INT_BUF_LEN - 1]++;
+    while (val) {
```
jserv, line 132: Add some comments and relevant materials.

visitorckw:
FWIW, I wrote a unit test to compare the results of the new function with the old one to verify correctness. Within the range [0, INT_MAX], the results of both functions are identical.

Testing code:
```
int x = 0;
char s1[INT_BUF_LEN];
char s2[INT_BUF_LEN];

while (1) {
    __str_base10(s1, x);
    new__str_base10(s2, x);
    if (strcmp(s1, s2)) {
        printf("Wrong answer when x = %d\n", x);
        return 0;
    }
    if (x == INT_MAX) break;
    x++;
}
```

jserv:

PR #149, opened 2024-08-11
Optimize trie operations by replacing tail recursion with iteration

visitorckw:
Replace tail-recursive calls in insert_trie() and find_trie() with an iterative loop to reduce function call overhead and avoid potential stack overflow issues. The logic remains functionally equivalent, improving performance and stability.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -87,25 +87,31 @@ char *elf_section;
  */
 int insert_trie(trie_t *trie, char *name, int funcs_index)
 {
-    char first_char = *name;
-    int fc = first_char;
-    if (!fc) {
-        if (!trie->index)
-            trie->index = funcs_index;
-        return trie->index;
+    char first_char;
+    int fc;
+
+    while (1) {
+        first_char = *name;
+        fc = first_char;
+        if (!fc) {
+            if (!trie->index)
+                trie->index = funcs_index;
+            return trie->index;
+        }
+        if (!trie->next[fc]) {
+            /* FIXME: The func_tries_idx variable may exceed the maximum number,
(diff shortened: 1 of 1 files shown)


src/globals.c
```diff
@@ -87,25 +87,31 @@ char *elf_section;
...
+            return trie->index;
+        }
+        if (!trie->next[fc]) {
+            /* FIXME: The func_tries_idx variable may exceed the maximum number,
+             * which can lead to a segmentation fault. This issue is affected by
+             * the number of functions and the length of their names. The proper
+             * way to handle this is to dynamically allocate a new element.
+             */
+            trie->next[fc] = func_tries_idx++;
+            for (int i = 0; i < 128; i++)
```
jserv, line 108: I am not sure if 128 is a proper range for such purpose. Can you determine?

visitorckw: In fact, the names of C functions can only contain a total of 63 characters: underscores, digits 0-9, and letters a-z and A-Z. Therefore, using 128 next entries for each node is indeed a waste of memory.

visitorckw: However, I think we might be able to address this issue in a later patch since it differs from the objective of this patch?

jserv: > However, I think we might be able to address this issue in a later patch since it differs from the objective of this patch?

up to you. We are experiencing performance issues during bootstrapping, and it would be beneficial to identify and resolve these to speed up the compilation process.

visitorckw: BTW, I think using a radix tree here could further improve memory savings. I can give it a try later.

jserv:
Thank @visitorckw for contributing! Let's move forward.

PR #151, opened 2024-09-14
Support octal literals

fennecJ:
As title, this PR basically implement support for octal literals in numeric constants and function parameters.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -612,22 +612,29 @@ void read_numeric_param(block_t *parent, basic_block_t *bb, int is_neg)
         is_neg = 1 - is_neg;
         i++;
     }
-    if ((token[0] == '0') && (token[1] == 'x')) { /* hexadecimal */
-        i = 2;
-        do {
-            c = token[i++];
-            if (is_digit(c))
-                c -= '0';
-            else {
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Read https://cbea.ms/git-commit/ carefully and refine the commits.

jserv:
You should extend the test suite for octal literals.

fennecJ:
I just added some test cases for octal literals, but I found that the current testing method has significant limitations. It seems like it can only test values between 0 and 255. Additionally, testing for negative numbers requires handling through printf. Should we consider improving our testing process to handle a broader range of cases more efficiently? What are your thoughts on this?

jserv:
> I just added some test cases for octal literals, but I found that the current testing method has significant limitations. It seems like it can only test values where the final result is between 0 and 255 (a byte) . Additionally, testing for negative numbers requires handling through printf. Should we consider improving our testing process to handle a broader range of cases more efficiently? What are your thoughts on this?

You can create a new issue to track the improvements of the test suite.

ChAoSUnItY (approved):

jserv:

PR #153, opened 2024-09-15
Check for invalid digits when parsing octal constants

DrXiao:
After introducing the pull request (#151 ), I use the following code to test the parser:

```c
/* test.c */
int main() {
    int a = 03, b = 01118, c = 091;
    printf("%d %d %d\n", a, b, c); 
    return 0;
}
```
```
$ qemu-arm test
3 592 73
```
Obviously, the octal numbers (`01118` and `091`) are invalid but the parser doesn't check whether they are valid or not.

Thus, the pull request improves the parser to enhance the validation check.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -634,7 +634,10 @@ void read_numeric_param(block_t *parent, basic_block_t *bb, int is_neg)
             } while (is_hex(token[i]));
         } else { /* octal */
             do {
-                c = token[i++] - '0';
+                c = token[i++];
+                if (c > '7')
+                    error("Invalid numeric constant");
+                c -= '0';
                 value = (value * 8) + c;
             } while (is_digit(token[i]));
         }
--- a/tests/driver.sh
+++ b/tests/driver.sh
@@ -62,6 +62,31 @@ function try_output() {
     try "$expected" "$expected_output" "$input"
 }
 
+# try_compile_error - test shecc with invalid C program
+# Usage:
+# - try_compile_error invalid_input_code
+# compile "invalid_input_code" with shecc so that shecc generates a
+# compilation error message.
+#
+# This function uses shecc to compile invalid code and obtains the exit
+# code returned by shecc. The exit code must be a non-zero value to
+# indicate that shecc has the ability to parse the invalid code and
+# output an error message.
+function try_compile_error() {
+    local input=$(cat)
+
+    local tmp_in="$(mktemp --suffix .c)"
+    local tmp_exe="$(mktemp)"
+    echo "$input" > "$tmp_in"
+    "$SHECC" -o "$tmp_exe" "$tmp_in"
+    local exit_code=$?
+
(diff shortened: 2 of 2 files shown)


src/parser.c
```diff
@@ -634,7 +634,10 @@ void read_numeric_param(block_t *parent, basic_block_t *bb, int is_neg)
             } while (is_hex(token[i]));
         } else { /* octal */
             do {
-                c = token[i++] - '0';
+                c = token[i++];
+                if (c > '7' || c < '0')
```
fennecJ, line 638: The check for `c < '0'` is redundant because the loop will exit as soon as `is_digit()` in line 642 returns false.   
For example, consider the case `int a = 0.; funct(a);`, where the ASCII value of `'.'` is less than `'0'`.   
On the first iteration, `c = token[i++]` assigns `c` the value `'0'` and increments `i` to 1.    
Then, the check in the while loop at line 642 `is_digit(token[i = 1])` will become `is_digit('.')` and return false, causing the loop to exit.   
As a result, the check for `c < '0'` is never triggered when the comming digit is less then '0'.

DrXiao: @fennecJ , yes, here doesn't need to check for `c < '0'`. Thank you for your feedback, I will improve it later.

ChAoSUnItY:
Please also add your code to the testsuite for code validation.

DrXiao:
The proposed changes let shecc generate a compilation error message when parsing invalid octal numbers.

However, originally, `tests/driver.sh` only compiles valid C programs and checks the exit code or standard output, and the test suite has never compiled any invalid C program before.

This is the first instance of adding an invalid C program, which includes some invalid octal constants, to  `test/driver.sh` to check whether (native) shecc generates the error message or not.

Thus, I will add a new function called `try_compile_error()` to the test suite, and the function will accept an invalid C program and then check if the exit code returned by shecc is a non-zero value. Eventually, the test suite will have the ability to test invalid C programs.

```bash
function try_compile_error() {
    local input=$(cat)

    local tmp_in="$(mktemp --suffix .c)"
    local tmp_exe="$(mktemp)"
    echo "$input" > "$tmp_in"
    "$SHECC" -o "$tmp_exe" "$tmp_in"
    local exit_code=$?

    if [ 0 == $exit_code ]; then
        echo "Error: compilation passed"
        exit 1
    fi
}


# ...

try_compile_error << EOF
int main() {
    int a = 03, b = 01118, c = 091;
    printf("%d %d %d\n", a, b, c); 
    return 0;
}
EOF
```

ChAoSUnItY (approved):

tests/driver.sh
```diff
@@ -62,6 +62,21 @@ function try_output() {
     try "$expected" "$expected_output" "$input"
 }
 
+function try_compile_error() {
```
jserv, line 65: Add comments here.

jserv:

PR #154, opened 2024-09-22
Skip symbol insertion for unreachable basic blocks

fennecJ:
Previously, the compiler would throw a segmentation fault if the compiled code contained a declaration after a return statement. For example:

int main() {
    return 0;
    int a = 5;
}

This occurred when the basic block (bb) for unreachable code was null, while the function add_symbol(basic_block_t *bb, var_t *var) did not check for null before attempting to insert a variable, leading to a segmentation fault.

This commit fixes the issue by skipping symbol insertion if the basic block is unreachable (null).


jserv:

PR #156, opened 2024-11-04
Implementation of  def-use chain

### LINKED ISSUE #155: Implementation of def-use-chain
While reviewing `ssa.c`, I came across the comment:

```c
/* TODO: simplify with def-use chain */
```
It made me wonder if the `compiler` already has a proper implementation of a def-use (DU) chain. The closest structure I found resembles a DU-chain, with some members inside the `var` struct:

```c
struct var {
    ...
    struct var *base;
    int subscript;
    struct var *subscripts[64];
    int subscripts_idx;
    struct insn *last_assign;
    rename_t rename;
    ...
};
```

From what I understand：
* `base` stores the original definition of the variable before renaming during SSA construction. 
* The `subscript` mirrors `base->rename.counter` to track how many times the variable has been renamed. 
* The `subscripts` array records the history of all previous variable versions before renaming. 
* The `last_assign` member seems to function like a use-def chain (i.e., the reverse of a DU chain), capturing the most recent assignment of the variable. 
* Finally, the `rename` member appears to track the renaming information for variables.

Could you confirm whether this design is meant to serve as a DU-chain or if further improvements are needed to implement one more explicitly? 

If a DU-chain still needs to be implemented, I propose the following approach:

For each `fn->bbs`, we introduce a new data structure, the DU-chain, which will be a hash map. The hash function will convert the string `var->var_name` and the rename count `var->subscript` into a hash value. Each DU-chain entry will store two pieces of data:

* `insn_t var_def`: the instruction defining the variable.
* `insn_list_t var_use`: a list of instructions where the variable is used.

Each DU-chain entry will be a linked list to handle potential hash collisions.

For every instruction:
* If it contains `rd` (a destination register), it will be inserted into the DU-chain based on the hashed key.
* If it contains `rs1`, `rs2` (source registers), the corresponding definition will be searched for in the DU-chain. If found, the instruction will be appended to the `var_use` list of the relevant DU entry. If not found, we will check if `var->is_global` is true. If not, an error will be raised for the use of an undeclared variable.

Please let me know if I’ve misunderstood anything or if this approach doesn't meet your expectations. I’m also open to any suggestions for improvement.

Thanks in advance!

nosba0957:
Implement a new structure for def-use chain called `uc_node_t`. `uc_node_t` is stored within `struct var`, includes `insn` element that records the instruction using the variable which the use chain belongs to.

Simplify CSE procedure with use chain information. We build the def-use chain information by `build_use_chain` function before the optimizing phase. In addtion, When the instructions are eliminated by CSE, we delete its use chain nodes from both variables(`rs1`, `rs2`) at the same time.

The CSE method has been modified. When the first pair of instructions is found, the use chain is utilized to search for identical ADD instructions. Subsequently, the next instruction is verified to ensure it match our target. After removing the instruction, utilize the use chain to find the next target instruction.

resolve #155

### DIFF
--- a/src/defs.h
+++ b/src/defs.h
@@ -157,6 +157,14 @@ struct ref_block_list {
 
 typedef struct ref_block_list ref_block_list_t;
 
+typedef struct insn insn_t;
+
+typedef struct use_chain_node {
+    insn_t *insn;
+    struct use_chain_node *next;
+    struct use_chain_node *prev;
+} use_chain_t;
+
 struct var {
     char type_name[MAX_TYPE_LEN];
     char var_name[MAX_VAR_LEN];
@@ -174,6 +182,8 @@ struct var {
     int subscripts_idx;
     rename_t rename;
     ref_block_list_t ref_block_list; /* blocks which kill variable */
+    use_chain_t *users_head;
+    use_chain_t *users_tail;
     struct insn *last_assign;
     int consumed;
     bool is_ternary_ret;
@@ -315,8 +325,6 @@ struct insn {
     phi_operand_t *phi_ops;
     char str[64];
 };
-
-typedef struct insn insn_t;
 
 typedef struct {
     insn_t *head;
--- a/src/ssa.c
+++ b/src/ssa.c
@@ -401,6 +401,54 @@ void build_rdf()
     free(args);
 }
 
+void use_chain_add_tail(insn_t *i, var_t *var)
+{
+    use_chain_t *u = calloc(1, sizeof(use_chain_t));
+    if (!u) {
+        printf("calloc failed\n");
+        abort();
+    }
+
+    u->insn = i;
+    if (!var->users_head)
+        var->users_head = u;
+    else
+        var->users_tail->next = u;
+    u->prev = var->users_tail;
+    var->users_tail = u;
+}
+
+void use_chain_delete(use_chain_t *u, var_t *var)
+{
+    if (u->prev)
+        u->prev->next = u->next;
+    else {
+        var->users_head = u->next;
+        u->next->prev = NULL;
+    }
+    if (u->next)
+        u->next->prev = u->prev;
+    else {
+        var->users_tail = u->prev;
+        u->prev->next = NULL;
+    }
+    free(u);
+}
+
+void use_chain_build()
+{
+    for (fn_t *fn = FUNC_LIST.head; fn; fn = fn->next) {
+        for (basic_block_t *bb = fn->bbs; bb; bb = bb->rpo_next) {
+            for (insn_t *i = bb->insn_list.head; i; i = i->next) {
+                if (i->rs1)
+                    use_chain_add_tail(i, i->rs1);
+                if (i->rs2)
+                    use_chain_add_tail(i, i->rs2);
+            }
+        }
+    }
+}
+
 bool var_check_killed(var_t *var, basic_block_t *bb)
 {
     for (int i = 0; i < bb->live_kill_idx; i++) {
@@ -1170,7 +1218,6 @@ void ssa_build(int dump_ir)
 }
 
 /* Common Subexpression Elimination (CSE) */
-/* TODO: simplify with def-use chain */
 /* TODO: release detached insns node */
 bool cse(insn_t *insn, basic_block_t *bb)
 {
@@ -1178,7 +1225,6 @@ bool cse(insn_t *insn, basic_block_t *bb)
         return false;
 
     insn_t *prev = insn->prev;
-
     if (!prev)
         return false;
     if (prev->opcode != OP_add)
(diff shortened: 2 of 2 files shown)


jserv (changes requested):
Rebase the latest `master` branch to resolve CI breakage.

src/defs.h
```diff
@@ -157,6 +157,14 @@ struct ref_block_list {
 
 typedef struct ref_block_list ref_block_list_t;
 
+typedef struct insn insn_t;
+
+typedef struct use_chain_node {
+    insn_t *insn;
+    struct use_chain_node *next;
+    struct use_chain_node *prev;
+} uc_node_t;
```
jserv, line 166: The naming scheme `uc_node_t` is not clear. Can you check [LLVM Programmer's Manual](https://llvm.org/docs/ProgrammersManual.html#iterating-over-def-use-use-def-chains) for refinement?

nosba0957: Sure, I will make adjustments.

src/defs.h
```diff
@@ -157,6 +157,14 @@ struct ref_block_list {
 
 typedef struct ref_block_list ref_block_list_t;
 
+typedef struct insn insn_t;
+
+typedef struct use_chain_node {
+    insn_t *insn;
+    struct use_chain_node *next;
+    struct use_chain_node *prev;
+} user_t;
```
jserv, line 166: Why not naming it `use_chain_t`?

jserv:
Can you provide test cases accordingly?

src/ssa.c
```diff
@@ -401,6 +401,49 @@ void build_rdf()
     free(args);
 }
 
+void users_add_tail(insn_t *i, var_t *var)
+{
+    user_t *u = calloc(1, sizeof(user_t));
```
fennecJ, line 406: Add handler when calloc failed (i.e. u == NULL)

src/defs.h
```diff
@@ -174,6 +182,8 @@ struct var {
     int subscripts_idx;
     rename_t rename;
     ref_block_list_t ref_block_list; /* blocks which kill variable */
+    user_t *users_head;
+    user_t *users_tail;
```
fennecJ, line 186: It seems like we can access users_tail directly through users_head->prev. Is the users_tail variable still necessary?

fennecJ (changes requested):

src/defs.h
fennecJ: I misunderstood; I thought you were implementing a circular doubly linked list, but it's actually just a doubly linked list without the circular characteristic.

src/defs.h
nosba0957: OK, I will make the modifications. I originally wanted to emphasize that this structure is for single node.

nosba0957:
> Can you provide test cases accordingly?

Do you mean adding test cases in driver.sh? But I don't think that would demonstrate the effect of CSE. Or should I provide the differences in assembly before and after implementing CSE?

src/ssa.c
nosba0957: OK, thank you for your reminder.

jserv:
> should I provide the differences in assembly before and after implementing CSE?

Yes, it makes sense.

src/ssa.c
```diff
@@ -401,6 +401,54 @@ void build_rdf()
     free(args);
 }
 
+void users_add_tail(insn_t *i, var_t *var)
```
jserv, line 404: Is `use_chain_add_tail` or `uc_add_tail` better for understanding?

src/ssa.c
```diff
@@ -401,6 +401,54 @@ void build_rdf()
...
+    if (u->next)
+        u->next->prev = u->prev;
+    else {
+        var->users_tail = u->prev;
+        u->prev->next = NULL;
+    }
+    free(u);
+}
+
+void build_users()
```
jserv, line 438: Enforce the consistent prefix. That is, `use_chain_build` or `uc_build`.

src/ssa.c
nosba0957: I think `use_chain_add_tail` is better; I'll change change it to that.

jserv:
Thank @nosba0957 for contributing!

fennecJ:
Apologies, I just reviewed the PR and was preparing my response.

The functionality looks correct, and the generated binary code matches the original implementation based on my offline tests. However, I have a concern about the simplification—it doesn’t seem to achieve the expected level of simplification. Nevertheless, I believe the DU-chain implementation can still be useful in the future for further optimizations.

Additionally, I think it’s important to properly clean up the allocated memory when constructing the DU-chain at the end of the compilation process.

jserv:
> I have a concern about the simplification—it doesn’t seem to achieve the expected level of simplification. Nevertheless, I believe the DU-chain implementation can still be useful in the future for further optimizations.
> Additionally, I think it’s important to properly clean up the allocated memory when constructing the DU-chain at the end of the compilation process.

No problem. Create another GitHub issue for tracking.

PR #158, opened 2024-11-07
Implement preprocessor directive #ifndef

ChAoSUnItY:
To prevent redefinition or serving other purposes, we sometimes need to negate the result of `defined`, however, currently shecc does support neither negation operator in macro scope, e.g. `!defined(SOME_MACRO)`, nor `#ifndef`. Thus, in this PR, we introduce preprocessor directive `#ifndef` to solve this issue without massively refactoring frontend codebase (since frontend refactoring is WIP).


jserv:

PR #161, opened 2024-11-08
Support empty alias and refine include guard tests

ChAoSUnItY:
In previous PR (#158), the include seems not to work correctly, thus, in this PR, I explicitly supports empty alias and refines previously added test case with additional include guard simulation test case.

### DIFF
--- a/tests/driver.sh
+++ b/tests/driver.sh
@@ -465,9 +465,49 @@ try_ 0 << EOF
 #else
 #define A 1
 #endif
+
+#ifndef A
+#define B 1
+#else
+#define B 0
+#endif
 int main()
 {
-    return A;
+    return A + B;
+}
+EOF
+
+# include guard test, simulates inclusion of a file named defs.h and global.c
+try_ 0 << EOF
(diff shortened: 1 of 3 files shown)


tests/driver.sh
```diff
@@ -465,9 +465,49 @@ try_ 0 << EOF
...
+#define B 0
+#endif
 int main()
 {
-    return A;
+    return A + B;
+}
+EOF
+
+# include gaurd test, simulates inclusion of a file named defs.h and global.c
```
jserv, line 480: Shall we add inclusion guard to all headers?

Also, fix typo. Should be "guard"

ChAoSUnItY: This would be a great enhancement for other editors without having to manually adjust tree-sitter-like plugins to have intellisense functionality. Should I include this modification in other PR or this one?

jserv: At the moment, shecc has only one header file, `src/defs.h`, so that it is straightforward to enforce inclusion guard.

jserv:

PR #167, opened 2024-11-23
Enhance code reusability for the compiler backends

DrXiao:
After observing ARM and RISC-V backends, I noticed that the instruction sequences of load and global load operations are very similar:
```diff
/* ARM backend */
if (ph2_ir->src0 > 4095) {
    emit(__movw(__AL, __r8, ph2_ir->src0));
    emit(__movt(__AL, __r8, ph2_ir->src0));
-   emit(__add_r(__AL, __r8, __sp, __r8));      /* OP_load */
+   emit(__add_r(__AL, __r8, __r12, __r8));     /* OP_global_load */
    emit(__lw(__AL, rd, __r8, 0));
} else
-   emit(__lw(__AL, rd, __sp, ph2_ir->src0));   /* OP_load */
+   emit(__lw(__AL, rd, __r12, ph2_ir->src0));  /* OP_global_load */
```
A similar case occurs for store and global store operations.

Thus, this pull request improves code reusability for the compiler backends.

PR #168, opened 2024-11-25
Refine code generation for address-of operations

DrXiao:
Upon further observation, I noticed that the instruction sequences of address-of and global address-of are highly similar and can be refined.

The proposed changes resemble my previous pull request (#167), but were discovered later, which is why I am submitting this pull request for improvement.


nosba0957 (approved):

jserv:

PR #169, opened 2024-12-01
Ensure correctness of formatted output conversion

DrXiao:
Consider the following code printing the several formatted integers:
```c
void print_conversion(int num) {
    printf("%#032x\n%#32x\n%032x\n%32x\n%#x\n%x\n", num, num, num, num, num, num);
    printf("%#012x\n%#12x\n%012x\n%12x\n", num, num, num, num);
    printf("%#032o\n%#32o\n%032o\n%32o\n%#o\n%o\n", num, num, num, num, num, num);
    printf("%#012o\n%#12o\n%012o\n%12o\n", num, num, num, num);
    printf("%032d\n%32d\n%d\n", num, num, num);
    printf("%012d\n%12d\n", num, num);
}
int main() {
    int a = 0xFF00CDE1, b = 0xFFFFF204, c = 65548, d = 0;
    print_conversion(a);
    print_conversion(b);
    print_conversion(c);
    print_conversion(d);
    return 0;
}
```
I found some incorrect results for certain outputs when using specifier, field width and alternate form:
### 1. Incorrect conversion from integers to hexadecimal strings
Consider a 32-bit integer `-16724511` (`0xFF00CDE1` in hexadecimal), if using `printf()` with the formatted string `"%#12x"`, it outputs `ffffff00cde1`, and the result has redundent `f` characters. However, because the integer is just 4 bytes, the result should be `    ff00cde1` containing 4 whitespace characters.

### 2. Incorrect conversion from integers to octal strings
As the previous point, consider the negative integer `-3580` (`0xFFFFF204` in hexadecimal), it produces `77777771004` when using `"%o"` to print this integer. 

Because the function accepts a 32-bit integer, the most significant digit of the octal should be obtained by the leftmost 2 bits only, but the original implementation doesn't consider this point and produce incorrect results.

Thus, the correct result is `37777771004`, which obtains `3` from the leftmost 2 bits of the integer.

### 3. Prepended characters may be placed in incorrect positions
When using field width and flag characters, it may output unexpected strings for certain cases.
`printf("%#12x\n", 0xff00cde1)` --> `"0x  ff00cde1\n"`
`printf("%#12o\n", 0x1000c) `     --> `"0     200014\n"`
`printf("%12d\n", -3580)`	       --> `"0000000-3580\n"`

## Conclusion
Through the given test code, the current built-in C library has been found to have the above issues, so this pull request resolves them and ensures correctness when using formatted output conversion.


lib/c.c
```diff
@@ -166,19 +166,29 @@ void __str_base10(char *pb, int val)
...
+    int c = INT_BUF_LEN - 1, v;
+    /*
+     * Because every 3 binary digits can be converted
+     * to 1 octal digit, here performs the conversion
+     * 10 times, derived from 32 divided by 3.
+     *
+     * Finally, the remaining 2 bits are processed after
+     * the loop.
+     * */
+    for (int i = 0; i < (sizeof(int) << 3) / 3; i++) {
```
ChAoSUnItY, line 178: Perhaps you can lift the rhs expression of condition expression out of loop to squeeze last bit of performance?

jserv:

PR #171, opened 2024-12-06
Support pointer data types for sizeof operator

ChAoSUnItY:
Currently `sizeof` is only capable of parsing basic data types like `int`, `char`, or user-defined data types like `struct`, `enum` etc. But not able to parsing pointer type, thus, in this PR, by introducing a temporary workaround parsing 0 or more asterisk tokens, we now partially support pointer type for `sizeof` operator (not including function pointer type yet).

Additionally, since the result given by `HOST_PTR_SIZE` in stage 0 and stage 1/2 are different, so I make another test driver named `driver-stage2.sh` for testing stage 2 shecc and adding corresponding `sizeof` test to verify the implementation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,10 +52,18 @@ $(OUT)/tests/%.elf: tests/%.c $(OUT)/$(STAGE0)
 	chmod +x $@ ; $(PRINTF) "Running $@ ...\n"
 	$(Q)$(TARGET_EXEC) $@ && $(call pass)
 
-check: $(TESTBINS) tests/driver.sh
-	tests/driver.sh
+check: check-stage0 check-stage2
+
+check-stage0: $(OUT)/$(STAGE0) $(TESTBINS) tests/driver.sh
+	$(VECHO) "  TEST STAGE 0\n"
+	tests/driver.sh 0
+
+check-stage2: $(OUT)/$(STAGE2) $(TESTBINS) tests/driver.sh
+	$(VECHO) "  TEST STAGE 2\n"
+	tests/driver.sh 2
 
 check-snapshots: $(OUT)/$(STAGE0) $(SNAPSHOTS) tests/check-snapshots.sh
+	$(VECHO) "  TEST SNAPSHOTS\n"
 	tests/check-snapshots.sh
 
 $(OUT)/%.o: %.c
--- a/README.md
+++ b/README.md
@@ -112,17 +112,27 @@ Verify that the emitted IRs are identical to the snapshots by specifying `check-
 $ make check-snapshots
 ```
 
-`shecc` comes with unit tests. To run the tests, give `check` as an argument:
+`shecc` comes with unit tests consist of stage 0, stage 2. To run these tests, give `check` as an argument:
 ```shell
 $ make check
 ```
 
 Reference output:
 ```
+  TEST STAGE 0
 ...
 int main(int argc, int argv) { exit(sizeof(char)); } => 1
 int main(int argc, int argv) { int a; a = 0; switch (3) { case 0: return 2; case 3: a = 10; break; case 1: return 0; } exit(a); } => 10
 int main(int argc, int argv) { int a; a = 0; switch (3) { case 0: return 2; default: a = 10; break; } exit(a); } => 10
+OK
+  TEST STAGE 2
+...
+int main(int argc, int argv) { exit(sizeof(char*)); }
+exit code => 4
+output => 
+int main(int argc, int argv) { exit(sizeof(int*)); }
+exit code => 4
+output => 
 OK
 ```
 
--- a/lib/c.c
+++ b/lib/c.c
@@ -14,6 +14,7 @@
 #define false 0
 
 #if defined(__arm__)
+#define __SIZEOF_POINTER__ 4
 #define __syscall_exit 1
 #define __syscall_read 3
 #define __syscall_write 4
@@ -23,6 +24,7 @@
 #define __syscall_munmap 91
 
 #elif defined(__riscv)
+#define __SIZEOF_POINTER__ 4
 #define __syscall_exit 93
 #define __syscall_read 63
 #define __syscall_write 64
--- a/src/parser.c
+++ b/src/parser.c
@@ -844,7 +844,11 @@ void read_expr_operand(block_t *parent, basic_block_t **bb)
         read_ternary_operation(parent, bb);
         lex_expect(T_close_bracket);
     } else if (lex_accept(T_sizeof)) {
+        /* TODO: Use more generalized type grammar parsing function to handle
+         * type reading
+         */
         char token[MAX_TYPE_LEN];
+        int ptr_cnt = 0;
 
         lex_expect(T_open_bracket);
         int find_type_flag = lex_accept(T_struct) ? 2 : 1;
@@ -853,9 +857,12 @@ void read_expr_operand(block_t *parent, basic_block_t **bb)
         if (!type)
             error("Unable to find type");
 
+        while (lex_accept(T_asterisk))
+            ptr_cnt++;
+
         ph1_ir = add_ph1_ir(OP_load_constant);
         vd = require_var(parent);
-        vd->init_val = type->size;
+        vd->init_val = ptr_cnt ? PTR_SIZE : type->size;
         strcpy(vd->var_name, gen_name());
         ph1_ir->dest = vd;
         opstack_push(vd);
--- a/src/reg-alloc.c
+++ b/src/reg-alloc.c
@@ -12,6 +12,18 @@
  * dead variable and does NOT wrtie it back to the stack.
  */
 
+/* Aligns size to nearest multiple of 4, this meets
+ * ARMv7's alignment requirement.
+ *
+ * This function should
+ * be called whenever handling with user-defined type's
+ * size.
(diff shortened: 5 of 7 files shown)


Makefile
```diff
@@ -52,12 +52,20 @@ $(OUT)/tests/%.elf: tests/%.c $(OUT)/$(STAGE0)
...
 
+check-stage2: $(OUT)/$(STAGE2) $(TESTBINS) tests/driver-stage2.sh
+	$(VECHO) "  TEST STAGE 2\n"
+	tests/driver-stage2.sh
+
 check-snapshots: $(OUT)/$(STAGE0) $(SNAPSHOTS) tests/check-snapshots.sh
+	$(VECHO) "  TEST SNAPSHOTS\n"
 	tests/check-snapshots.sh
 
+check-all: $(OUT)/$(STAGE2) $(TESTBINS) check check-stage2
```
jserv, line 67: Don't name it `check-all`. Instead, consolidate the existing `check` for stage1 and stage2.

tests/driver-stage2.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
```
jserv, line 1: I have no idea why `tests/driver-stage2.sh` exits.

tests/driver-stage2.sh
```diff
@@ -0,0 +1,110 @@
...
+    local expected="$1"
+    local input="$2"
+    items "$expected" "exit($input);"
+}
+
+# sizeof
+expr 4 "sizeof(int)";
+expr 1 "sizeof(char)";
+# sizeof pointers
+expr 4 "sizeof(void*)";
```
DrXiao, line 106: According to your proposed changes, I think here can also adds the following cases:
```c
/* size of a pointer to pointer to integer or other basic data types */
sizeof(int **);
```
```c
typedef struct {
    int a;
    int b;
} test_struct;

 /* size of a pointer to a structure */
sizeof(test_struct *);
```

ChAoSUnItY:
I'm unable to reproduce broken pipe failure on my machine, and the snapshots I checked which are generated by `tests/update-snapshots.sh` are same as the original one.

DrXiao:
Because this pull request involves size of pointers, I would like to ask a question.

In the header file (`def.h`), we can notice that there are two macros called `PTR_SIZE`([Line 47](https://github.com/sysprog21/shecc/blob/18e005dbe21b323dc1559195388e02e7ff101bba/src/defs.h#L47)) and `HOST_PTR_SIZE`  ([Line 59 ~ 64](https://github.com/sysprog21/shecc/blob/18e005dbe21b323dc1559195388e02e7ff101bba/src/defs.h#L59)). Obviously, both of them mean the size of a pointer, but I cannot understand their use and difference all the time, especially the later one (`HOST_PTR_SIZE`).

Thus, I hope to obtain an explanation about them through this opportunity.

ChAoSUnItY:
> In the header file (def.h), we can notice that there are two macros called PTR_SIZE ([Line 47](https://github.com/sysprog21/shecc/blob/18e005dbe21b323dc1559195388e02e7ff101bba/src/defs.h#L47)) and HOST_PTR_SIZE ([Line 59 ~ 64](https://github.com/sysprog21/shecc/blob/18e005dbe21b323dc1559195388e02e7ff101bba/src/defs.h#L59)). Obviously, both of them mean the size of a pointer, but I cannot understand their use and difference all the time, especially the later one (HOST_PTR_SIZE).

The `HOST_PTR_SIZE` is determined by the running host, while `PTR_SIZE` is supposed to be 4, which is capable of handling the arch we're targeting to. The only usages of `HOST_PTR_SIZE` currently are `memcpy` in `src/ssa.c`.

Which I later realized value of `sizeof(type*)` should not be `HOST_PTR_SIZE`, instead, it should be `PTR_SIZE`.

DrXiao:
@ChAoSUnItY , after fetching your latest changes, I try to build shecc and check the snapshots:
```shell
$ make distclean && make ARCH=arm && tests/check-snapshots.sh 
...
  SHECC	out/shecc-stage1.elf
  SHECC	out/shecc-stage2.elf
```
```shell
$ make distclean && make ARCH=riscv && tests/check-snapshots.sh 
...
  SHECC	out/shecc-stage1.elf
  SHECC	out/shecc-stage2.elf
```
It seems that the snapshots differs when shecc uses RISC-V backend. Hope this information helps you.

ChAoSUnItY:
Ok, I found the cause, apparently previously shecc's CI won't check snapshot on rv32 but only on arm32:

https://github.com/sysprog21/shecc/blob/18e005dbe21b323dc1559195388e02e7ff101bba/Makefile#L55-L59

https://github.com/sysprog21/shecc/blob/18e005dbe21b323dc1559195388e02e7ff101bba/.github/workflows/main.yml#L6-L27

So currently the workaround is just not to run check-snapshot on rv32.

Should we later address this issue in anther PR? @jserv

jserv:
> Ok, I found the cause, apparently previously shecc's CI won't check snapshot on rv32 but only on arm32:

That was a fault. Let's fix it in another pull request.

tests/driver-stage2.sh
jserv: Is it possible to stick to single `tests/driver.sh` for validation?

ChAoSUnItY: Yes, it would be possible. Just need to plug in 1 one argument to specify which stage's executable is being used with testing.

ChAoSUnItY: I have tested `tests/driver.sh` with stage 2, but I have found several problems:
1. The compilation speed is significantly slower than stage 0.
2. Stage 2 compiler would failed to compile these test suites
https://github.com/sysprog21/shecc/blob/18e005dbe21b323dc1559195388e02e7ff101bba/tests/driver.sh#L955-L985

DrXiao: > Stage 2 compiler would failed to compile these test suites

I use the following test code to test shecc first.
```c
int main() 
{ 
    printf("%d %d\n", -2147483648 / 1, -2147483648 % 1); 
    return 0;  
} 
```
Using current shecc to compile the code would result in a Bus Error, and I believe the potential cause is `__stmdb` and `__ldm` in the Arm backend, which operate on memory to store or load multiple words, .

If temporarily removing these instructions from `arm-codegen.c`, shecc will compile the test code successfully.
```diff
    case OP_div:
    case OP_mod:
        ...
        interm = __r8;
        /* div/mod emulation */
        /* Preserve the values of the dividend and divisor */
-       emit(__stmdb(__AL, 1, __sp, (1 << rn) | (1 << rm)));
        /* Obtain absolute values of the dividend and divisor */
        emit(__srl_amt(__AL, 0, arith_rs, __r8, rn, 31));
        ...
        emit(__mov_r(__AL, __r9, rn));
-       emit(__ldm(__AL, 1, __sp, (1 << rn) | (1 << rm)));
        emit(__mov_r(__AL, rd, interm));
        ...
        return;
```
---

That is, the behavior of `__stmdb` and `__ldm` may lead to unaligned memory access, thereby triggering a `Bus Error`.

Therefore, I think improving the current Arm bakcend could be a solution to resolve the issue.

reference: [ARM Cortex-A Series Programmer's Guide for ARMv7-A](https://developer.arm.com/documentation/den0013/d/Porting/Alignment)
> However, load and store multiple instructions (LDM and STM) and load and store double-word (LDRD or STRD) must be aligned to at least a word boundary.

ChAoSUnItY: Currently I have resolved ARM stack misalignment by ensure `__stmdb` and `__ldm` pushes 4 registers at a time to keep it aligned.

But now the RISCV backend will stuck, but I found that current implementation does not check the formerly mentioned test suite edge case scenario: `INT_MIN` could not be approximated due to missing equality check, which leads the program keep shifting the approximated value after reaching `INT_MIN` then causes the approximated value always be 0 afterwards.

By adding additional equality test while approximating nearest power of 2 value, the program would work as expected.

ChAoSUnItY:
I will wrap up some additional changes introduced besides `sizeof` operator implementation enhancement:

Previously, while using stage 2 compiler to compile test suites in `tests/driver.sh`, the arm32 suffers bus error, and the rv32 suffers infinite loop.

The former one was caused by improper stack alignment, while the later one was caused by missing edge case equality check, which is `INT_MIN`. Thus, I have added `stack_align` function in `reg_alloc.c` to ensure every stack increment is properly aligned, and I have added a branching instruction for the edge case.

The change of total stack size of shecc itself has increased from 9225 to 9228.

src/arm-codegen.c
```diff
@@ -336,7 +336,9 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
         }
         interm = __r8;
         /* div/mod emulation */
-        /* Preserve the values of the dividend and divisor */
+        /* Preserve the values of the dividend and divisor. This also keeps
```
DrXiao, line 339: Because the latest changes preserve the original use of `__stmdb` and `__ldm`, I think the comment doesn't need to be changed.

jserv:
I defer to @vacantron for confirmation.

vacantron (approved):
LGTM

jserv:

PR #175, opened 2025-01-09
Fix bad logical-or implementation

DrXiao:
Because the current shecc still generates bitwise-or instructions for logical-or operations, the proposed changes improve both the frontend and backend so that it generates branch instructions for logical-or operations.

The new implementation obeys short-circuit principle; that is, once the first operand is true, the second operand is not evaluated. Also, the proposed changes add some test cases the validate the correctness.

### DIFF
--- a/src/defs.h
+++ b/src/defs.h
@@ -23,7 +23,7 @@
 #define MAX_FUNC_TRIES 2160
 #define MAX_BLOCKS 2048
 #define MAX_TYPES 64
-#define MAX_IR_INSTR 40000
+#define MAX_IR_INSTR 50000
 #define MAX_BB_PRED 128
 #define MAX_BB_DOM_SUCC 64
 #define MAX_BB_RDOM_SUCC 256
@@ -190,7 +190,7 @@ struct var {
     struct insn *last_assign;
     int consumed;
     bool is_ternary_ret;
-    bool is_log_and_ret;
+    bool is_logical_ret;
     bool is_const; /* whether a constant representaion or not */
 };
 
--- a/src/parser.c
+++ b/src/parser.c
@@ -300,11 +300,9 @@ int read_constant_infix_expr(int precedence)
             lhs = lhs != rhs;
             break;
         case OP_log_and:
-            /* TODO: Short-circuit evaluation */
             lhs = lhs && rhs;
             break;
         case OP_log_or:
-            /* TODO: Short-circuit evaluation */
             lhs = lhs || rhs;
             break;
         default:
@@ -692,9 +690,10 @@ void read_char_param(block_t *parent, basic_block_t *bb)
     add_insn(parent, bb, OP_load_constant, ph1_ir->dest, NULL, NULL, 0, NULL);
 }
 
-void read_log_and_operation(block_t *parent,
-                            basic_block_t **bb,
-                            char *label_else);
+void read_logical(opcode_t op,
+                  block_t *parent,
+                  basic_block_t **bb,
+                  char *label_shared);
 void read_ternary_operation(block_t *parent, basic_block_t **bb);
 void read_func_parameters(block_t *parent, basic_block_t **bb)
 {
@@ -1013,10 +1012,16 @@ void read_expr_operand(block_t *parent, basic_block_t **bb)
     }
 }
 
-void finalize_log_and(block_t *parent,
+void finalize_logical(opcode_t op,
+                      block_t *parent,
                       basic_block_t **bb,
-                      char *label_else,
-                      basic_block_t *else_bb);
+                      char *label_shared,
+                      basic_block_t *shared_bb);
+
+bool is_logical(opcode_t op)
+{
+    return op == OP_log_and || op == OP_log_or;
+}
 
 void read_expr(block_t *parent, basic_block_t **bb)
 {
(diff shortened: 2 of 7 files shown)


src/defs.h
```diff
@@ -190,7 +190,7 @@ struct var {
     struct insn *last_assign;
     int consumed;
     bool is_ternary_ret;
-    bool is_log_and_ret;
+    bool is_log_op_ret;
```
jserv, line 193: How about renaming to `is_logical_ret`?

src/parser.c
```diff
@@ -692,9 +690,10 @@ void read_char_param(block_t *parent, basic_block_t *bb)
     add_insn(parent, bb, OP_load_constant, ph1_ir->dest, NULL, NULL, 0, NULL);
 }
 
-void read_log_and_operation(block_t *parent,
-                            basic_block_t **bb,
-                            char *label_else);
+void read_log_op(opcode_t op,
```
jserv, line 693: Ditto. How about `read_logical`?

tests/driver.sh
```diff
@@ -1272,4 +1272,261 @@ int main()
...
+    printf("(a || b) && (c || d) = %d\n", (a || b) && (c || d));
+    printf("a && (b || c || d) = %d\n", a && (b || c || d));
+    printf("a * 0 && (b || c || d) = %d\n", a * 0 && (b || c || d));
+    printf("a * 2 && (b || c || d) = %d\n", a * 2 && (b || c || d));
+    printf("a && (b * 0 || c || d * 0) = %d\n", a && (b * 0 || c || d * 0));
+
+    printf("func(a) && func(b) && func(c) = %d\n", func(a) && func(b) && func(c));
+    printf("func(a) - a && func(b) && func(c) = %d\n", func(a) - a && func(b) && func(c));
+    printf("func(a) && func(b) && func(c) + 1 = %d\n", func(a) - a && func(b) && func(c) + 1);
+    printf("func(c) || func(a) && func(d) = %d\n", func(c) || func(a) || func(d));
```
fennecJ, line 1521: Is the difference between the printed statement and the function evaluation intentional?

tests/driver.sh
```diff
@@ -1272,4 +1272,261 @@ int main()
...
+    printf("func(c) || func(c) && func(d) = %d\n", func(c) || func(c) || func(d));
```
fennecJ, line 1522: Ditto, printed statement has different operation compare to function evaluation

tests/driver.sh
```diff
@@ -1272,4 +1272,261 @@ int main()
...
+    printf("(func(c) || func(c)) && func(d + 100) = %d\n", (func(c) || func(c)) && func(d + 100));
+    printf("func(c) || func(a) && func(d) = %d\n", func(c) || func(a) && func(d));
+    printf("func(a) && (func(d) || func(c)) = %d\n", func(a) && (func(d) || func(c)));
+    printf("func(a) && (func(d) || func(c)) = %d\n", func(a) * 2 && (func(d) || func(c)));
```
fennecJ, line 1526: Ditto, printed statement has different operation compare to function evaluation

DrXiao: Thanks. I will fix them.

src/parser.c
```diff
@@ -1440,105 +1564,190 @@ void read_log_and_operation(block_t *parent,
...
+         * |                     T |
+         * |                       |
+         * assign                  assign
+         * false                   true
+         * (shared_bb,             ('then_next')
+         *  'else_bb')             |
+         * |                       |
+         * |                       |
+         * |                       |
+         * ----------------------> ('end')
```
jserv, line 1618: Can you refine the ASCII art? I didn't get the idea.

DrXiao: I will think a better ASCII art to illustrate how it works here.

fennecJ: I think @DrXiao want to express a short-circuit flow here for the logic a && b && c. If the evaluation (test) of a is false, the process will directly follow the left-hand path to the assign false corner at the bottom left, bypassing the evaluations of b and c.
Perhaps adding some arrows to indicate the direction of the flow could make the ASCII diagram clearer?

DrXiao:
```c
/*  bb1
 * +------------+
 * | teq  a, #0 |      False
 * | bne  bb2   | --------------->
 * | b    bb5   |                 |
 * +------------+                 |
 *        |                       |
 *        | True                  |
 *  bb2   |                       |
 * +------------+                 |
 * | teq  b, #0 |      False      |
 * | bne  bb3   | --------------->
 * | b    bb5   |                 |
 * +------------+                 |
 *        |                       |
 *        | True                  |
 *  bb3   |                       |    bb5
 * +------------+      False      |   +------------+
 * | teq  c, #0 | ------------------> | load  0    |
 * | bne  bb4   |                     | b     bb6  |
 * | b    bb5   |                     +------------+
 * +------------+                           |
 *        |                                 |
 *        | True                            |
 *  bb4   |                                 |
 * +-----------+                            |
 * | load  1   |                            |
 * | b     bb6 |                            |
 * +-----------+                            |
 *        |            bb6                  |
 *        |           +--------+            |
 *        --------->  |        |  <----------
 *                    |        |
 *                    +--------+
 * 
 * bb3, bb4, bb5, and bb6 will be 'then', 'then_next', 'else_bb' 
 * and 'end', respectively.
 * */
```
Here is my current draft of the latest ASCII art. The purpose of adding the comment is to explain the use of the variables `then`, `then_next`, `else_bb` and `end`.

For finalizing the logical operation, it should use the mentioned variables to add some instructions to these basic blocks and properly connect them to each other.

jserv: > Here is my current draft of the latest ASCII art. The purpose of adding the comment is to explain the use of the variables `then`, `then_next`, `else_bb` and `end`.

Can you rotate the above "figure" 90 degrees clockwise, so that it would look more compact?

DrXiao:
```c
/* For example: a && b
 * 
 *
 *  bb1                 bb2                bb3
 * +-----------+       +-----------+       +---------+
 * | teq a, #0 | True  | teq b, #0 | True  | ldr 1   |
 * | bne bb2   | ----> | bne bb3   | ----> | b   bb5 |
 * | b   bb4   |       | b   bb4   |       +---------+
 * +-----------+       +-----------+         |
 *      |                |                   |
 *      | False          | False             |
 *      |                |                   |
 *      |              +---------+         +--------+
 *      -------------> | ldr 0   | ------> |        |
 *                     | b   bb5 |         |        |
 *                     +---------+         +--------+
 *                      bb4                 bb5
 * 
 * bb2, bb3, bb4, and bb5 will be 'then', 'then_next', 
 * 'else_bb' and 'end', respectively.
 * */
```
Does this updated figure look better?

Because the original example (`a && b && c`) may make the comment too long after rotating the figure, I decided to use the simplest case (`a && b`) to explain how it works here.

jserv: It makes sense.

jserv:

PR #177, opened 2025-01-16
Refine comment and division operator tokenization logic

ChAoSUnItY:
This refactor removes original unreachable error branch on `/` character-related tokenizations, and also adds C++-style comment (which is introduced in C99 specification) support. Additionally, the C-style comment tokenizatin is refined with proper detection on unenclosed case.


jserv:

PR #178, opened 2025-01-17
Support all assignment operators

ChAoSUnItY:
Previously, only `+=`, `-=`, `|=`, and `&=` assignment operators are supported, in this commit, we've support the remaining assignment operators, which are `*=`, `/=`, `%=`, `<<=`, `>>=`, and `^=`.

Additionally, this commit also refactors some assignment operations with the operators introduced in this commit.


jserv:

PR #179, opened 2025-01-18
Replace func trie with hashmap

ChAoSUnItY:
Previously, trie implementation is not consistent, mainly because of using index to point the referencing `func_t` to `FUNCS`, additionally, it lacks of dynamic allocation which might cause segmentation fault and results more technical debt to debug on either `FUNCS` or `FUNCS_TRIE`. Thus, in this PR, we can resolve this issue by introducing a dynamic hashmap.

Current implementation is using [FNV-1a](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV-1a_hash) hashing algorithm (32-bit edition to be precise), and due to lack of unsigned integer implementation, hashing result ranges from `0` to `2,147,483,647`.

Notice that current implementation may suffer from lookup issue when the function amount keeps increasing since current hashmap implementation doesn't offer rehashing based on load factor (which ideally, 0.75 would be best and currently shecc does not support floating number).

This also enables us to refactor more structures later with hashmap implementation in shecc.

<details>

<summary>Benchmark for ./tests/hello.c compilation</summary>

### Before

```shell
Command being timed: "./out/shecc tests/hello.c"
        User time (seconds): 0.00
        System time (seconds): 0.02
        Percent of CPU this job got: 76%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.03
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 52112
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 12220
        Voluntary context switches: 0
        Involuntary context switches: 0
        Swaps: 0
        File system inputs: 0
        File system outputs: 32
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

### After

```shell
Command being timed: "./out/shecc tests/hello.c"
        User time (seconds): 0.00
        System time (seconds): 0.02
        Percent of CPU this job got: 71%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.03
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 49916
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 12224
        Voluntary context switches: 1
        Involuntary context switches: 0
        Swaps: 0
        File system inputs: 8
        File system outputs: 32
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

</details> 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request replaces the trie structure for function lookups with a dynamic hashmap, improving memory management and addressing segmentation faults. The new implementation uses the FNV-1a hashing algorithm but currently lacks dynamic resizing, which may impact performance as function count increases. Overall, it simplifies function storage and sets the stage for future enhancements.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -15,18 +15,13 @@ block_list_t BLOCKS;
 macro_t *MACROS;
 int macros_idx = 0;
 
-/* the first element is reserved for global scope */
-func_t *FUNCS;
-int funcs_idx = 1;
-
-/* FUNC_TRIES is used to improve the performance of the find_func function.
- * Instead of searching through all functions and comparing their names, we can
- * utilize the trie data structure to search for existing functions efficiently.
- * The index starts from 1 because the first trie node represents an empty input
- * string, and it is not possible to record a function with an empty name.
+/* FUNCS_MAP is used to integrate function storing and boost lookup
+ * performance, currently it uses FNV-1a hash function to hash function
+ * name. The bucket size defaults to MAX_FUNCS. Ideally, it should be a small
+ * number, but due to lack of rehashing implementation, to prevent collision,
+ * we have to initially create large amount of buckets.
  */
-trie_t *FUNC_TRIES;
-int func_tries_idx = 1;
+hashmap_t *FUNCS_MAP;
 
 type_t *TYPES;
 int types_idx = 0;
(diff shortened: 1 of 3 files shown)


jserv:
@visitorckw, can you comment this?

src/globals.c
```diff
@@ -72,72 +61,159 @@ char *elf_strtab;
...
- * find_trie() - search the index of the function name in the trie
- * @trie: A pointer to the trie where the name will be searched.
- * @name: The name to be searched.
+ * create_hashmap() - creates a hashmap on heap.
+ * @size: The initial bucket size of hashmap.
  *
- * Return: The index of the pointer to the func_t.
+ * @returns: The pointer of created hashmap.
+ */
+hashmap_t *create_hashmap(int size)
```
jserv, line 102: Rename to `hashmap_create` for consistency.

src/globals.c
```diff
@@ -72,72 +61,159 @@ char *elf_strtab;
...
+/**
+ * new_node() - creates a hashmap node on heap.
+ * @key: The key of node. Must not be NULL.
+ * @val: The value of node. Could be NULL.
  *
- * 0 - the name not found.
- * otherwise - the index of the founded index in the trie array.
+ * @returns: The pointer of created node.
+ */
+node_t *new_node(char *key, void *val)
```
jserv, line 121: Ditto. Enforce consistent naming scheme.

visitorckw:
Looks good as is.

However, as mentioned, a large number of functions may cause excessive collisions and slow down performance. For smaller function counts, the default 512 buckets might be overkill. Therefore, a radix tree with dynamic memory allocation could still be a method worth exploring in the future.

ChAoSUnItY:
I'm concerning that dynamic memory allocation at this moment is not reliable and potentially flawed, I've attempted to implement rehashing algorithm before, but on stage 2 the compilation will fail, while the GCC and stage 1 are fine.

src/globals.c
```diff
@@ -72,72 +61,159 @@ char *elf_strtab;
...
 
 /**
- * insert_trie() - Inserts a new element into the trie structure.
- * @trie: A pointer to the trie where the name will be inserted.
- * @name: The name to be inserted into the trie.
- * @funcs_index: The index of the pointer to the func_t. The index is recorded
- *     in a 1-indexed format. Because the first element of 'FUNCS' has been
- *     reserved, there is no need to shift it.
- * Return: The index of the pointer to the func_t.
+ * hash_fnv1a() - hashes a string with FNV-1a hash function.
```
jserv, line 64: Improve the comments for hash function evaluation and considerations.

jserv: Since the optimizer lacks of the ability to inline functions, you can simply move the body of `hash_fnv1a` into `hash_index`.

src/globals.c
```diff
@@ -72,72 +61,159 @@ char *elf_strtab;
...
+/**
+ * contains_hashmap() - checks if the key-value pair entry exists
+ * from given key.
+ * @map: The hashmap to be looked up. Must no be NULL.
+ * @key: The key string. May be NULL.
+ *
+ * @returns: The look up result, if the key-value pair entry
+ * exists, then returns true, false otherwise.
+ */
+bool contains_hashmap(hashmap_t *map, char *key)
```
jserv, line 189: Rename to `hashmap_contains`.

src/globals.c
```diff
@@ -72,72 +61,159 @@ char *elf_strtab;
...
+{
+    return hashmap_get(map, key) != NULL;
+}
+
+/**
+ * free_hashmap() - frees the hashmap, this also frees key-value pair
+ * entry's value.
+ * @map: The hashmap to be looked up. Must no be NULL.
+ */
+void free_hashmap(hashmap_t *map)
```
jserv, line 199: Ditto. Enforce consistent naming scheme.

jserv:
Refine the message "This also enables us to refactor more structures later with hashmap implementation in shecc." as following:
> This also allows for future refactoring of additional structures using a hashmap implementation.

You don't have to mention the project name unless it does impact something you are referring.

src/globals.c
```diff
@@ -12,18 +12,7 @@ block_list_t BLOCKS;
...
-
-/* FUNC_TRIES is used to improve the performance of the find_func function.
- * Instead of searching through all functions and comparing their names, we can
- * utilize the trie data structure to search for existing functions efficiently.
- * The index starts from 1 because the first trie node represents an empty input
- * string, and it is not possible to record a function with an empty name.
- */
-trie_t *FUNC_TRIES;
-int func_tries_idx = 1;
+hashmap_t *FUNCS_MAP;
```
jserv, line 15: Add comments.

src/globals.c
```diff
@@ -72,72 +61,149 @@ char *elf_strtab;
...
 
 /**
- * insert_trie() - Inserts a new element into the trie structure.
- * @trie: A pointer to the trie where the name will be inserted.
- * @name: The name to be inserted into the trie.
- * @funcs_index: The index of the pointer to the func_t. The index is recorded
- *     in a 1-indexed format. Because the first element of 'FUNCS' has been
- *     reserved, there is no need to shift it.
- * Return: The index of the pointer to the func_t.
+ * hash_index_hashmap() - hashses a string with FNV-1a hash function
```
jserv, line 64: Refine the comments.

src/globals.c
```diff
@@ -72,72 +61,149 @@ char *elf_strtab;
...
+{
+    int index = hash_index(map->size, key);
+    hashmap_node_t *cur = map->buckets[index];
+
+    while (cur) {
+        if (!strcmp(cur->key, key))
+            return cur->val;
+
+        cur = cur->next;
+    }
```
jserv, line 165: Rewrite use `for` iteration for short.

src/globals.c
```diff
@@ -72,72 +61,149 @@ char *elf_strtab;
...
+
+        while (cur) {
+            next = cur->next;
+            free(cur->key);
+            free(cur->val);
+            /* FIXME: Remove this if-clause will cause double free error */
+            if (cur != map->buckets[0])
+                free(cur);
+            cur = next;
+        }
```
jserv, line 202: Ditto. Use `for` iteration for short.

src/globals.c
```diff
@@ -72,72 +61,149 @@ char *elf_strtab;
...
+ * from given key.
+ * @map: The hashmap to be looked up. Must no be NULL.
+ * @key: The key string. May be NULL.
+ *
+ * @returns: The look up result, if the key-value pair entry
+ * exists, then returns true, false otherwise.
+ */
+bool hashmap_contains(hashmap_t *map, char *key)
+{
+    return hashmap_get(map, key) != NULL;
```
jserv, line 181: Remove `!= NULL` as it does explicitly.

src/globals.c
```diff
@@ -75,72 +70,143 @@ char *elf_strtab;
...
-             * way to handle this is to dynamically allocate a new element.
-             */
-            trie->next[fc] = func_tries_idx++;
-            for (int i = 0; i < 128; i++)
-                FUNC_TRIES[trie->next[fc]].next[i] = 0;
-            FUNC_TRIES[trie->next[fc]].index = 0;
-        }
-        trie = &FUNC_TRIES[trie->next[fc]];
-        name++;
+int hash_index(int size, char *key)
```
jserv, line 82: Use `hashmap_` prefix as it would be used only in hashmap implementation so far.

src/globals.c
```diff
@@ -75,72 +70,143 @@ char *elf_strtab;
...
-        }
-        trie = &FUNC_TRIES[trie->next[fc]];
-        name++;
+int hashmap_hash_index(int size, char *key)
+{
+    int hash = 0x811c9dc5, mask;
+
+    for (; *key; key++) {
+        hash ^= *key;
+        hash *= 0x01000193;
```
visitorckw, line 88: The multiplication here may cause overflow, leading to undefined behavior. Signed integer overflow is undefined, while unsigned integer overflow is not. Since shecc currently lacks support for unsigned integers, we might consider adding it to address this issue.

ChAoSUnItY: I think we can just simply adds `unsigned` type at this moment to simplify the effort of new type? This way, unsigned can still uses current signed's arithmetic algorithm (due to the fact that they both based on 2's complement), since in ARMv7 and RISC-V 32bit assembly, signed overflow is well-defined, the only difference would be intrepretation of most-significant bit.

jserv: > should we postpone this hashmap implementation?

You can simply convert this pull request to draft state.

ChAoSUnItY: I think this pull request should be implemented as soon as possible, the reason is that I'm currently working on [`type_t` refactor](https://github.com/ChAoSUnItY/shecc/tree/refactor/types), but I have encountered the weird `free(): invalid pointer` issue when adding functions in `globals.c`, which I assume the reason is that the function trie cannot hold more than certain numbers of functions and probably function name's length could also contribute to this issue, these 2 factors and the flaw is already described here:
https://github.com/sysprog21/shecc/blob/09bb9180e5ac46b7c2df03104c6fcbbb56db9988/src/globals.c#L103-L109**
But after cherry-picked this branch without any changes to function structures, the issue immediately gone.

One possible solution towards this is to add `-fwrapv` compilation flag to `gcc` to instruct compiler to wrap signed integer overflow result according to the 2's compliment representation, this ensures defined behavior at least when compiling with `gcc`. Meanwhile in shecc, it's fine at this moment since both ARM 32bit and RISC-V 32bit assembly wraps the overflow value according to the 2's compliment representation as well.

vacantron: Could enlarging limitation in `refs.h` temporary fix this problem? I remember the trie count is almost reach the limitation in the last change.

This workaround could be remove after applying this patch.

ChAoSUnItY: Changing macro `MAX_FUNC_TRIES` in `defs.h` to 3000 does the trick.

visitorckw: > I think this pull request should be implemented as soon as possible, the reason is that I'm currently working on [`type_t` refactor](https://github.com/ChAoSUnItY/shecc/tree/refactor/types), but I have encountered the weird `free(): invalid pointer` issue when adding functions in `globals.c`, which I assume the reason is that the function trie cannot hold more than certain numbers of functions and probably function name's length could also contribute to this issue, these 2 factors and the flaw is already described here:
> 
> https://github.com/sysprog21/shecc/blob/09bb9180e5ac46b7c2df03104c6fcbbb56db9988/src/globals.c#L103-L109
> 
I don't necessarily oppose this PR. However, if the issue is that `MAX_FUNC_TRIES` is too small, causing an out-of-bounds array access, it seems unrelated to switching to a hash table instead of a trie. A hash table can also use an array, and a trie can use dynamic memory allocation. This feels more like adding a new feature unrelated to fixing the bug itself. But I'm fine if we decide to switch to a hash table as a workaround.

ChAoSUnItY: The major reason that I would like to replace trie is that some errors are not straightforward to be realized, in this case, it generates `free(): invalid pointer`, which isn't that obvious to do with insufficient trie size and is not friendly to new-coming contributors in my opinion.

src/globals.c
```diff
@@ -75,72 +70,143 @@ char *elf_strtab;
...
+ * and converts into usable hashmap index. The range of returned
+ * hashmap index is ranged from "(0 ~ 2,147,483,647) mod size" due to
+ * lack of unsigned integer implementation.
+ * @size: The size of map. Must not be negative or 0.
+ * @key: The key string. May be NULL.
  *
- * If the function has been inserted, the return value is the index of the
- * function in FUNCS. Otherwise, the return value is the value of the parameter
- * @funcs_index.
+ * @returns: The usable hashmap index.
```
jserv, line 80: Use `Return` instead of `@returns`.

src/globals.c
```diff
@@ -75,72 +70,143 @@ char *elf_strtab;
...
+{
+    int hash = 0x811c9dc5, mask;
+
+    for (; *key; key++) {
+        hash ^= *key;
+        hash *= 0x01000193;
+    }
+
+    mask = hash >> 31;
+    return ((hash ^ mask) - mask) % size;
```
jserv, line 92: Can you ensure that `size` is always power of 2? If so, we can eliminate the modulo operation.

src/globals.c
```diff
@@ -75,72 +70,185 @@ char *elf_strtab;
...
+        return NULL;
+    }
+
+    map->size = round_up_pow2(size);
+    map->buckets = malloc(map->size * sizeof(hashmap_node_t *));
+
+    if (!map->buckets) {
+        printf("Failed to allocate buckets in hashmap_t\n");
+        return NULL;
+    }
```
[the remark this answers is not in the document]
ChAoSUnItY: Duplicated review

src/globals.c
```diff
@@ -75,72 +70,185 @@ char *elf_strtab;
...
+{
+    int hash = 0x811c9dc5, mask;
+
+    for (; *key; key++) {
+        hash ^= *key;
+        hash *= 0x01000193;
     }
+
+    mask = hash >> 31;
+    return ((hash ^ mask) - mask) & (size - 1);
```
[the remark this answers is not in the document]
ChAoSUnItY: This issue is previously discussed to be ignored at this moment.

src/globals.c
```diff
@@ -75,72 +70,185 @@ char *elf_strtab;
...
+{
+    for (int i = 0; i < map->size; i++) {
+        for (hashmap_node_t *cur = map->buckets[i], *next; cur; cur = next) {
+            next = cur->next;
+            free(cur->key);
+            free(cur->val);
+            /* FIXME: Remove this if-clause will cause double free error */
+            if (cur != map->buckets[0])
+                free(cur);
+            cur = next;
```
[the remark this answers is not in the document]
ChAoSUnItY: This free logic design is meant to prevent invalid pointer free issue caused by shecc.

src/globals.c
```diff
@@ -15,18 +15,13 @@ block_list_t BLOCKS;
...
-/* the first element is reserved for global scope */
-func_t *FUNCS;
-int funcs_idx = 1;
-
-/* FUNC_TRIES is used to improve the performance of the find_func function.
- * Instead of searching through all functions and comparing their names, we can
- * utilize the trie data structure to search for existing functions efficiently.
- * The index starts from 1 because the first trie node represents an empty input
- * string, and it is not possible to record a function with an empty name.
+/* FUNCS_MAP is used to integerate function storing and boost lookup
```
jserv, line 18: Typo: integerate

src/globals.c
```diff
@@ -15,18 +15,13 @@ block_list_t BLOCKS;
...
+ * performance, currently it uses FNV-1a hash function to hash function
```
jserv, line 19: Fix grammar mistakes.

jserv:

PR #182, opened 2025-03-20
Fix dereference behavior on mixed subscript and arrow / dot operators

### LINKED ISSUE #164: On-heap struct pointer assignment corrupts memory
Considering the following example minimum arena implementation:

```c
#include <stdlib.h>

typedef enum {
    A,
    B
} type_t;

typedef struct {
    type_t typ;
} data_t;

typedef struct {
    int cap;
    int size;
    data_t *data;
} arena_t;

arena_t *arena_init(int cap) {
    arena_t *arena = malloc(sizeof(arena_t));
    arena->cap = cap;
    arena->size = 0;
    arena->data = calloc(cap, sizeof(data_t));
    return arena;
}

data_t *arena_alloc(arena_t *arena) {
    if (arena->size + 1 >= arena->cap)
        return NULL;

    return &arena->data[arena->size++];
}

void arena_free(arena_t *arena) {
    free(arena->data);
    free(arena);
}

int main() {
    arena_t *arena = arena_init(1024);
    data_t *data = arena_alloc(arena);
    printf("%d\n", arena->data);
    data->typ = B;
    printf("%d\n", arena->data);
    arena_free(arena);

    return 0;
}
```

I expect that it would output 2 numbers that refers to same memory address where `arena->data` resides, however, using shecc results in the second number is different to the first one, and same as the integer value of enum `B` previously assigned.

Using gcc to compile this program would have expected result however.

### LINKED ISSUE #165: Invalid char pointer deference behavior
Considering the following code:
```c
#include <stdlib.h>
#include <string.h>

char a[100];

typedef struct {
    char *raw;
} data_t;

void init_data(data_t *data, char *raw) {
    data->raw = raw;
}

int main() {
    strcpy(a, "DATA");
    data_t *data = malloc(sizeof(data_t));
    init_data(data, a);
    
    char *raw = data->raw;
    printf("%c\n", a[0]);
    printf("%c\n", raw[0]);
    printf("%c\n", data->raw[0]);
    
    
    free(data);

    return 0;
}
```

Using gcc to compile the program and run it, it would output:
```
D
D
D
```

But using shecc, it would output:
```
D
D
�
```

This happens because of the incorrect deference assembly code.

### LINKED ISSUE #181: Incorrect double free behavior
Using the code below to initialize buckets for hashmap and frees hashmap:

```c
#include <stdlib.h>

typedef struct hashmap_node {
    char *key;
    void *val;
    struct hashmap_node *next;
} hashmap_node_t;

typedef struct {
    int size;
    hashmap_node_t **buckets;
} hashmap_t;

hashmap_t *hashmap_create(int size)
{
    hashmap_t *map = malloc(sizeof(hashmap_t));
    map->size = size;
    map->buckets = malloc(size * sizeof(hashmap_node_t *));

    for (int i = 0; i < map->size; i++)
        map->buckets[i] = 0;

    return map;
}

void hashmap_free(hashmap_t *map)
{
    for (int i = 0; i < map->size; i++) {
        for (hashmap_node_t *cur = map->buckets[i], *next; cur; cur = next) {
            next = cur->next;
            free(cur->key);
            free(cur->val);
            free(cur);
            cur = next;
        }
    }

    free(map->buckets);
    free(map);
}

int main()
{
    hashmap_t *map = hashmap_create(16);
    map->buckets[0] = malloc(sizeof(hashmap_node_t)); // Simulates put at first bucket
    map->buckets[0]->key = calloc(1, sizeof(char));
    map->buckets[0]->val = malloc(sizeof(int));
    map->buckets[0]->next = NULL;
    hashmap_free(map);
    return 0;
}
```

Compile with gcc will exit normally, while compile with shecc will cause double free, which is abnormal.

ChAoSUnItY:
- Fix #165.
- Fix #181, as it's already introduced into upstream source code by #179.
- Fix #164.
- This change could hopefully allow us to safely operate more dynamic data strctures without having troubles. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request fixes a double free error in the globals file and enhances type safety in the parser by changing variable types from int to bool. It also introduces new tests for dynamic data structures, improving the robustness of the codebase and addressing issues #165, #181, and #164.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -750,7 +750,7 @@ void read_lvalue(lvalue_t *lvalue,
                  var_t *var,
                  block_t *parent,
                  basic_block_t **bb,
-                 int eval,
+                 bool eval,
                  opcode_t op);
 
 /* Maintain a stack of expression values and operators, depending on next
@@ -805,7 +805,7 @@ void read_expr_operand(block_t *parent, basic_block_t **bb)
 
         lex_peek(T_identifier, token);
         var_t *var = find_var(token, parent);
-        read_lvalue(&lvalue, var, parent, bb, 0, OP_generic);
+        read_lvalue(&lvalue, var, parent, bb, false, OP_generic);
 
         if (!lvalue.is_reference) {
             ph1_ir = add_ph1_ir(OP_address_of);
@@ -825,7 +825,7 @@ void read_expr_operand(block_t *parent, basic_block_t **bb)
         int open_bracket = lex_accept(T_open_bracket);
         lex_peek(T_identifier, token);
         var_t *var = find_var(token, parent);
-        read_lvalue(&lvalue, var, parent, bb, 1, OP_generic);
+        read_lvalue(&lvalue, var, parent, bb, true, OP_generic);
         if (open_bracket)
             lex_expect(T_close_bracket);
 
@@ -962,7 +962,7 @@ void read_expr_operand(block_t *parent, basic_block_t **bb)
         } else if (var) {
             /* evalue lvalue expression */
             lvalue_t lvalue;
-            read_lvalue(&lvalue, var, parent, bb, 1, prefix_op);
+            read_lvalue(&lvalue, var, parent, bb, true, prefix_op);
 
             /* is it an indirect call with function pointer? */
             if (lex_peek(T_open_bracket, NULL)) {
@@ -1234,13 +1234,13 @@ void read_lvalue(lvalue_t *lvalue,
                  var_t *var,
                  block_t *parent,
                  basic_block_t **bb,
-                 int eval,
+                 bool eval,
                  opcode_t prefix_op)
 {
     ph1_ir_t *ph1_ir;
     var_t *vd;
-    int is_address_got = 0;
-    int is_member = 0;
+    bool is_address_got = false;
+    bool is_member = false;
 
     /* already peeked and have the variable */
     lex_expect(T_identifier);
@@ -1260,6 +1260,23 @@ void read_lvalue(lvalue_t *lvalue,
     while (lex_peek(T_open_square, NULL) || lex_peek(T_arrow, NULL) ||
            lex_peek(T_dot, NULL)) {
         if (lex_accept(T_open_square)) {
+            /* if subscripted member's is not yet resolved, dereference to
+             * resolve base address.
+             * e.g., dereference of "->" in "data->raw[0]" would be performed
+             * here.
+             */
+            if (lvalue->is_reference && lvalue->is_ptr && is_member) {
+                ph1_ir = add_ph1_ir(OP_read);
+                ph1_ir->src0 = opstack_pop();
+                vd = require_var(parent);
+                strcpy(vd->var_name, gen_name());
+                ph1_ir->dest = vd;
+                opstack_push(vd);
+                ph1_ir->size = 4;
+                add_insn(parent, *bb, OP_read, ph1_ir->dest, ph1_ir->src0, NULL,
+                         ph1_ir->size, NULL);
+            }
+
             /* var must be either a pointer or an array of some type */
             if (var->is_ptr == 0 && var->array_size == 0)
                 error("Cannot apply square operator to non-pointer");
@@ -1303,8 +1320,8 @@ void read_lvalue(lvalue_t *lvalue,
                      ph1_ir->src1, 0, NULL);
 
             lex_expect(T_close_square);
-            is_address_got = 1;
-            is_member = 1;
+            is_address_got = true;
+            is_member = true;
             lvalue->is_reference = true;
         } else {
             char token[MAX_ID_LEN];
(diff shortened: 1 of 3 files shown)


tests/driver.sh
```diff
@@ -433,6 +433,38 @@ int main() {
 }
 EOF
 
+# Mixed subscription and arrow operator,
```
DrXiao, line 436: I recommend rewrite it as `Mixed subscript and arrow operators`
(In C99 6.5.2.1, `[ ]` is called  the "subscript operator".)

src/parser.c
```diff
@@ -1260,6 +1260,18 @@ void read_lvalue(lvalue_t *lvalue,
     while (lex_peek(T_open_square, NULL) || lex_peek(T_arrow, NULL) ||
            lex_peek(T_dot, NULL)) {
         if (lex_accept(T_open_square)) {
+            if (lvalue->is_reference && lvalue->is_ptr && is_member) {
```
DrXiao, line 1263: How about writing some comments here?

src/parser.c
```diff
@@ -1260,6 +1260,20 @@ void read_lvalue(lvalue_t *lvalue,
     while (lex_peek(T_open_square, NULL) || lex_peek(T_arrow, NULL) ||
            lex_peek(T_dot, NULL)) {
         if (lex_accept(T_open_square)) {
+            /* if previous operator requires deference, deference here, then
```
jserv, line 1263: Should be "dereference."

src/parser.c
```diff
@@ -1260,6 +1260,20 @@ void read_lvalue(lvalue_t *lvalue,
...
+             * offset later */
```
jserv, line 1264: Improve the descriptions.

ChAoSUnItY: I just tested additional expressions e.g. "data.raw[0]" to better conclude this patch's effect, apperently this does also fix deference behavior on mixed subscript and dot operator (which also produced incorrect code before).

jserv (changes requested):
Clarify the use of "deference" and "dereference" in the wording.

src/parser.c
```diff
@@ -1260,6 +1260,22 @@ void read_lvalue(lvalue_t *lvalue,
     while (lex_peek(T_open_square, NULL) || lex_peek(T_arrow, NULL) ||
            lex_peek(T_dot, NULL)) {
         if (lex_accept(T_open_square)) {
+            /* if subscripted member's is not yet resolved, deference to resolve
```
jserv, line 1263: There should be a comma (`,`) appearing right after "e.g.". That is, `e.g.,`.

See https://www.scribbr.com/definitions/for-example-abbreviation/

tests/driver.sh
```diff
@@ -433,6 +433,49 @@ int main() {
...
+
+int main() {
+    strcpy(a, "DATA");
+    data_t *data = malloc(sizeof(data_t));
+    data->raw = a;
+    data_t data2;
+    data2.raw = a;
+    char *raw = data->raw;
+    char *raw2 = data2.raw;
+    // mixed arrow / dot with subscript operators dereference
```
DrXiao, line 456: Although shecc supports `//`-style comments, I think it is better to write single-line comments in C89 style (`/* ... */`), even though this is just test code.

reference: [shecc/CONTRIBUTING.md](https://github.com/sysprog21/shecc/blob/master/CONTRIBUTING.md)

src/parser.c
```diff
@@ -1260,6 +1260,22 @@ void read_lvalue(lvalue_t *lvalue,
     while (lex_peek(T_open_square, NULL) || lex_peek(T_arrow, NULL) ||
            lex_peek(T_dot, NULL)) {
         if (lex_accept(T_open_square)) {
+            /* if subscripted member's is not yet resolved, dereference to
```
DrXiao, line 1263: How about separating the example from the explanation of the implementation?

That is, rewrite it as follows:
```c
/* If subscripted member's ....
 * ... base address.
 *
 * e.g., dereference of "->" ...
 * would be performed here.
 */
```

DrXiao (changes requested):
I notice that the title of this pull request still uses "deference" in the wording. Fixing it to "dereference" is better and prevents misunderstandings.

Additionally, the git commit title is `Fix incorrect deref behavior`, and I believe it could be improved with a clearer summary.

src/parser.c
```diff
@@ -1260,6 +1260,23 @@ void read_lvalue(lvalue_t *lvalue,
...
+             * resolve base addres.
```
DrXiao, line 1264: typo: addres

DrXiao (commented):
Except for the typo in the comment that needs to be fixed, please adjust the git commit message.

After fixing the typo and refining the git commit message, I think the proposed changes look good.

jserv (changes requested):
Append the `Close #?` where `?` stands for the issue you are managing to close at the end of git commit messages.

jserv:

PR #183, opened 2025-03-25
Eliminate arithmetic instructions based on properties

ChAoSUnItY:
This patch eliminates arithmetic instructions (`add`, `sub`, and `mul`, in particular) when the criterion is matched. Including:
- Identity property for addition.
- Zero and identity property for multiplication, these are effective when doing pointer offset.
- Subtraction will be replaced based on operand, e.g., "s - 0" will be replaced with "s"; "0 - s" will be replaced with "-s".

Notice that `insn_fusion`'s return value has changed from `void` to `bool` to preserve ability to indicate whether instruction has been optimized or not. 

## Analysis of performance statistics

Before: `hyperfine --warmup 3 './out/shecc ./src/main.c' './out/shecc-stage1.elf ./src/main.c'`
```
Benchmark 1: ./out/shecc ./src/main.c
  Time (mean ± σ):     966.5 ms ± 185.1 ms    [User: 277.6 ms, System: 616.3 ms]
  Range (min … max):   755.6 ms … 1237.0 ms    10 runs

Benchmark 2: ./out/shecc-stage1.elf ./src/main.c
  Time (mean ± σ):      8.060 s ±  0.452 s    [User: 4.461 s, System: 3.036 s]
  Range (min … max):    7.495 s …  8.931 s    10 runs
```

After: `hyperfine --warmup 3 './out/shecc ./src/main.c' './out/shecc-stage1.elf ./src/main.c'`
```
Benchmark 1: ./out/shecc ./src/main.c
  Time (mean ± σ):      1.126 s ±  0.428 s    [User: 0.324 s, System: 0.713 s]
  Range (min … max):    0.801 s …  2.283 s    10 runs

Benchmark 2: ./out/shecc-stage1.elf ./src/main.c
  Time (mean ± σ):      8.166 s ±  0.263 s    [User: 4.356 s, System: 3.181 s]
  Range (min … max):    7.936 s …  8.760 s    10 runs
```

I assume the minor performance regression is caused by introduction of more condition checking on every instructions.

### DIFF
--- a/src/peephole.c
+++ b/src/peephole.c
@@ -31,23 +31,114 @@ bool is_fusible_insn(ph2_ir_t *ph2_ir)
     }
 }
 
-void insn_fusion(ph2_ir_t *ph2_ir)
+bool insn_fusion(ph2_ir_t *ph2_ir)
 {
     ph2_ir_t *next = ph2_ir->next;
     if (!next)
-        return;
+        return false;
 
     if (next->op == OP_assign) {
-        /* eliminate {ALU rn, rs1, rs2; mv rd, rn;} */
-        if (!is_fusible_insn(ph2_ir))
-            return;
-        if (ph2_ir->dest == next->src0) {
+        if (is_fusible_insn(ph2_ir) && ph2_ir->dest == next->src0) {
+            /* eliminates:
+             * {ALU rn, rs1, rs2; mv rd, rn;}
+             * reduces to:
+             * {ALU rd, rs1, rs2;}
+             */
             ph2_ir->dest = next->dest;
             ph2_ir->next = next->next;
-            return;
+            return true;
         }
     }
-    /* other insn fusions */
+
+    if (ph2_ir->op == OP_load_constant && ph2_ir->src0 == 0) {
+        if (next->op == OP_add &&
+            (ph2_ir->dest == next->src0 || ph2_ir->dest == next->src1)) {
+            /* eliminates:
+             * {li rn, 0; add rd, rs1, rn;} or
+             * {li rn, 0; add rd, rn, rs1;}
+             * reduces to:
+             * {mv rd, rs1;}, based on identity property of addition
+             */
+            /* Determine the non-zero source operand */
+            int non_zero_src =
+                (ph2_ir->dest == next->src0) ? next->src1 : next->src0;
+
+            /* Transform instruction sequence from addition with zero to move */
+            ph2_ir->op = OP_assign;
+            ph2_ir->src0 = non_zero_src;
+            ph2_ir->dest = next->dest;
+            ph2_ir->next = next->next;
+            return true;
+        }
+
+        if (next->op == OP_sub) {
+            if (ph2_ir->dest == next->src1) {
+                /* eliminates:
+                 * {li rn, 0; sub rd, rs1, rn;}
+                 * reduces to:
+                 * {mv rd, rs1;}
+                 */
+                ph2_ir->op = OP_assign;
+                ph2_ir->src0 = next->src0;
+                ph2_ir->dest = next->dest;
+                ph2_ir->next = next->next;
+                return true;
+            }
+
+            if (ph2_ir->dest == next->src0) {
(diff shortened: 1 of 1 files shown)


src/peephole.c
```diff
@@ -31,23 +31,105 @@ bool is_fusible_insn(ph2_ir_t *ph2_ir)
...
+            ph2_ir->src0 = ph2_ir->dest == next->src0 ? next->src1 : next->src0;
+            ph2_ir->dest = next->dest;
+            ph2_ir->next = next->next;
+            return true;
+        }
+    }
+
     /* other insn fusions */
+
+    return false;
```
jserv, line 132: Refine the comment.

src/peephole.c
```diff
@@ -31,23 +31,105 @@ bool is_fusible_insn(ph2_ir_t *ph2_ir)
...
     if (next->op == OP_assign) {
-        /* eliminate {ALU rn, rs1, rs2; mv rd, rn;} */
-        if (!is_fusible_insn(ph2_ir))
-            return;
-        if (ph2_ir->dest == next->src0) {
+        if (is_fusible_insn(ph2_ir) && ph2_ir->dest == next->src0) {
+            /* eliminates:
+             * {ALU rn, rs1, rs2; mv rd, rn;}
+             * reduces to:
+             * {ALU rd, rs1, s2;}
```
DrXiao, line 45: Should it be `{ALU rd, rs1, rs2;}` here ?

ChAoSUnItY: Yes, it's a typo, I'll patch it in next commit.

src/peephole.c
```diff
@@ -63,6 +149,7 @@ void peephole()
                     ir->next = next->next;
                     continue;
                 }
+
```
DrXiao, line 152: Is this blank line necessary?

ChAoSUnItY: No. I'll remove it later.

jserv (changes requested):
Instead of referring the instructions as "math," classify them in the field of "arithmetic instructions."

src/peephole.c
```diff
@@ -31,23 +31,109 @@ bool is_fusible_insn(ph2_ir_t *ph2_ir)
...
+             * {li rn, 0; add rd, rn, rs1;}
+             * reduces to:
+             * {mv rd, rs1;}, based on identity property of addition
+             */
+            ph2_ir->op = OP_assign;
+            ph2_ir->src0 = ph2_ir->dest == next->src0 ? next->src1 : next->src0;
+            ph2_ir->dest = next->dest;
+            ph2_ir->next = next->next;
+            return true;
+        }
```
jserv, line 67: Consider the changes below:
```c
if (ph2_ir->op == OP_load_constant && ph2_ir->src0 == 0) {
    if (next->op == OP_add &&
        (ph2_ir->dest == next->src0 || ph2_ir->dest == next->src1)) {
        /* Determine the non-zero source operand */
        int non_zero_src = (ph2_ir->dest == next->src0) ? next->src1 : next->src0;

        /* Transform instruction sequence from addition with zero to move */
        ph2_ir->op = OP_assign; 
        ph2_ir->src0 = non_zero_src;
        ph2_ir->dest = next->dest;
        ph2_ir->next = next->next;
        return true;
    }
}
```

jserv:
The results generated by `hyperfine --warmup 3 'out/shecc src/main.c'` looks weird. Can you utilize tools such as [uftrace](https://github.com/namhyung/uftrace) to figure out the slowdown?

ChAoSUnItY:
The result given by `uftrace` is contrary to the result generated by `hyperfine`, the optimized one is faster then the original one.
image

vacantron (approved):
LGTM. Did you test stage-1 on armv7 machine?

ChAoSUnItY:
No, and I don't have any armv7 / riscv machine to run tests. I am currently using qemu to bootstrap and benchmark locally, which may not be accurate.

jserv:

PR #187, opened 2025-04-07
Correct the return value of fputc function

DrXiao:
According to the man page, the return value of `fputc()` should be either the character written as an `unsigned char` cast to an `int` or `EOF` on error, but the current implementation doesn't follow the specification.

Therefore, the proposed changes correct the implementation to return the ASCII code of the written character or -1 if the output fails, thereby matching the description in the man page.

Additionally, the necessary test cases are also added to validate the correctness. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request corrects the fputc function implementation to return the ASCII code of the written character or -1 on error, as per specifications. It also adds comprehensive test cases to validate the function's behavior, improving reliability and testing coverage.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>


lib/c.c
```diff
@@ -496,10 +496,10 @@ char *fgets(char *str, int n, FILE *stream)
 
 int fputc(int c, FILE *stream)
 {
-    char buf[1];
+    int buf[2];
```
jserv, line 499: Reduced from 2 to 1 since we only need to write 1 character.

tests/driver.sh
```diff
@@ -1611,4 +1611,36 @@ int main()
...
+	int c = fputc('a', stdout);
+	printf("written = %c", c);
+	return 0;
+}
+EOF
+
+try_output 1 "" << EOF
+#define stdout 1
+
+int main()
```
jserv, line 1638: Ditto.

jserv (changes requested):
Rebase the latest `master` branch and resolve the conflicts.

jserv:

PR #191, opened 2025-04-21
Overhaul `hashmap_t` functionalities

### LINKED ISSUE #190: Rewrite contract of `hashmap_t` to be able to store `NULL` as value
Currently, using `NULL` as value in hashmap would cause 2 unexpected results, including:
- `hashmap_contains` returns false whenever the bucket node that stores `NULL` exist or not.
- `hashmap_free` results in segmentation fault due to unchecked bucket node pointer value.

This would be useful as we can treat `hashmap_t` as an alternative of hash set by storing `NULL` as value.

ChAoSUnItY:
This patch fixes constracts of `hashmap_t`-related function, including:

- `hashmap_contains`: Instead of delegates to result returned by `hashmap_get`, now shares similar logic to avoid bucket node value being `NULL` and treat as false.
- ~~`hashmap_free`: Now respects bucket node value's pointer, if it's already `NULL`, then hashmap won't free it.~~

Additionally, now `hashmap_t `is capable of rehashing its underlying bucket node array when load factor reaches **75%**.

Closes #190. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the `hashmap_t` functionalities by implementing a rehashing mechanism triggered at a 75% load factor, refining the `hashmap_contains` logic to avoid incorrect NULL returns, and updating `hashmap_free` for better memory management. These improvements optimize the performance and reliability of the hashmap, ensuring a more robust implementation.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -231,26 +231,24 @@ int round_up_pow2(int v)
  *
  * Return: The pointer of created hashmap.
  */
-hashmap_t *hashmap_create(int size)
+hashmap_t *hashmap_create(int cap)
 {
     hashmap_t *map = malloc(sizeof(hashmap_t));
 
     if (!map) {
-        printf("Failed to allocate hashmap_t with size %d\n", size);
+        printf("Failed to allocate hashmap_t with capacity %d\n", cap);
         return NULL;
     }
 
-    map->size = round_up_pow2(size);
-    map->buckets = malloc(map->size * sizeof(hashmap_node_t *));
+    map->size = 0;
+    map->cap = round_up_pow2(cap);
+    map->buckets = calloc(map->cap, sizeof(hashmap_node_t *));
 
     if (!map->buckets) {
         printf("Failed to allocate buckets in hashmap_t\n");
         free(map);
         return NULL;
     }
-
-    for (int i = 0; i < map->size; i++)
-        map->buckets[i] = 0;
 
     return map;
 }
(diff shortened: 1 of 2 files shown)


src/globals.c
```diff
@@ -368,7 +424,8 @@ void hashmap_free(hashmap_t *map)
         for (hashmap_node_t *cur = map->buckets[i], *next; cur; cur = next) {
             next = cur->next;
             free(cur->key);
-            free(cur->val);
+            if (cur->val)
+                free(cur->val);
```
jserv, line 428: Don't do this.

jserv: Quote from [free(3)](https://man7.org/linux/man-pages/man3/free.3p.html):
> If `ptr` is a null pointer, no action shall occur.

src/globals.c
```diff
@@ -304,42 +343,59 @@ void hashmap_put(hashmap_t *map, char *key, void *val)
...
     } else {
         while (cur->next)
             cur = cur->next;
-        cur->next = hashmap_node_new(key, val);
+        cur->next = new_node;
     }
 
-    /* TODO: Rehash if size exceeds size * load factor */
+    map->size++;
+    if ((map->cap >> 2) * 3 <= map->size)
```
DrXiao, line 359: How about replacing `(map->cap >> 2) * 3` with  `((map->cap >> 1) + (map->cap >> 2))` to compute an equivalent result while avoiding a multiply instruction?

src/globals.c
```diff
@@ -304,42 +343,60 @@ void hashmap_put(hashmap_t *map, char *key, void *val)
...
         while (cur->next)
             cur = cur->next;
-        cur->next = hashmap_node_new(key, val);
+        cur->next = new_node;
     }
 
-    /* TODO: Rehash if size exceeds size * load factor */
+    map->size++;
+    /* Check if size of map exceeds load factor 75% (or 3/4 of capacity) */
+    if ((map->cap >> 2) * (map->cap >> 1) <= map->size)
```
DrXiao, line 360: It should be `(map->cap >> 2) + (map->cap >> 1) <= map->size`.

ChAoSUnItY: My bad, I'll push another patch.

DrXiao (approved):

jserv:

PR #192, opened 2025-04-26
Correct the behavior and return value of snprintf()

DrXiao:
Recently, I tested the following C code in the test suite:
```c
int main() {
    char buffer[20];
    int written = snprintf(buffer, 0, "Number: %d", -37);
    printf("%s", buffer);
    return written;
}
```
If using the current shecc to test, the exit code is `0`. However, the exit code is `11` if using GCC to validate it.

Due to this difference, I read C99 7.19.6.5 to understand the specification of `snprintf()` and noticed the following description in that section:
> *  If n is zero, nothing is written.
> * Output characters beyond the n-1st are discarded rather than being written to the array, and a null character is written at the end of the characters actually written into the array.
> * The snprintf function returns the number of characters that would have been written
had n been sufficiently large.

After observing the current implementation of `snprintf()` in the built-in C library and ran the related test cases again, I found that the behavior and return value of `snprintf()` did not match the above description.

Therefore, the proposed changes improve the implementation of `snprintf()` to comply with the specification, and the related test cases are also adjusted to verify the correct behavior. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the snprintf function in the C library to comply with the C99 standard, particularly for zero size parameters. Modifications ensure correct return values and behavior, along with the addition of new test cases for validation, improving reliability across compilers.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 2 - The changes are straightforward and well-documented, making the review process relatively easy.
</div>


tests/driver.sh
```diff
@@ -1647,6 +1647,12 @@ int main() {
 }
 EOF
 
+# The following cases validate the behavior and return value of
+# snprintf(). Notice that you should refer to the specification,
+# such as C99 7.19.6.5, to understand the behavior of snprintf().
```
jserv, line 1652: Instead of saying "you should refer to," quote the essential contents from C standards if you want to emphasize.

DrXiao:
Consider the following code:
```
int main()
{
    char buffer[20];
    for (int i = 0; i < 19; i++)
        buffer[i] = '0';
    buffer[19] = 0;

    int written = snprintf(buffer, 10, "Number: %06d", -35337);

    for (int i = 0; i < 20; i++)
        printf(" %x", buffer[i]);
    printf("\n");
    return written;
}
```
When using shecc with the proposed changes and GCC to test, it writes at most 10 bytes.
```
 4e 75 6d 62 65 72 3a 20 2d 0 30 30 30 30 30 30 30 30 30 0
```
However, using the original shecc to compile and execute, it writes more than 10 bytes.
```
 4e 75 6d 62 65 72 3a 20 2d 0 35 33 33 37 30 30 30 30 30 0
```

This test code also shows that the original implementation of `snprintf()` is faulty. I will add more test cases to validate and refine the comments in the test suite.

lib/c.c
```diff
@@ -210,16 +210,55 @@ void __str_base16(char *pb, int val)
     }
 }
 
-int __format(char *buffer,
+typedef struct {
+    char *buf;
+    int n;
+} fmtbuf_t;
```
vacantron, line 216: Maybe store the number of written bytes or the original `buf` that we can simply return the string length by them instead of doing the calculation stuff of `len`?

DrXiao: I have defined a new member called `len` in the structure to store the return value for the functions in the `printf()` family.

lib/c.c
```diff
@@ -210,15 +210,86 @@ void __str_base16(char *pb, int val)
...
+ * - Output characters beyond the n-1st are discarded rather than
+ *   being written to the array.
+ * - The snprintf function returns the number of characters that would
+ *   have been written had n been sufficiently large, not counting
+ *   the terminating null character, or a negative value if an encoding
+ *   error occurred.
+ *
+ * Therefore, the following code defines a structure called fmtbuf_t
+ * to implement formatted output conversion for the functions in the
+ * printf() family.
```
jserv, line 228: Shorten the quotation, making the comments lean and still useful.

vacantron (approved):
LGTM

ChAoSUnItY (approved):

jserv:

PR #195, opened 2025-04-28
Integrate function-related structures

ChAoSUnItY:
Since the introduction of SSA middle end, there exists 2 different usages of function-related structures that is heavily coupled, but used under different scenarios. This causes unnecessary confusions to anyone who tries to understand it, and makes maintenance costly.

In this patch, `func_t` and `fn_t` are now integrated into `func_t`, and related variables are renamed into `func` instead of `fd` or `fn` to keep naming consistency and avoid confusions to previous structures.

Additionally, few comments are refined, and unused label definitions removed to keep codebase tidy.


src/defs.h
```diff
@@ -524,27 +507,34 @@ struct ref_block {
     struct ref_block *next;
 };
 
-/* TODO: integrate func_t into fn_t */
-struct fn {
+struct func {
```
jserv, line 510: Add comments to describe the usage of this struct.

jserv:

PR #196, opened 2025-05-06
Remove deprecated phase 1 IR (`ph1_ir_t`)

ChAoSUnItY:
Phase 1 IR is previously served as a solution for bridging C frontend to backend (elf code generation), which becomes obsolete after the introduction of SSA middle-end and the introduction of new phase 1 IR called `insn_t`, after this, `ph1_ir_t` is only used for IR dumping purpose. In this patch, `ph1_ir_t` and related structures & functionalities are removed and `insn_t` now covers the original usage of `ph1_ir_t`.

Additionally, IR dumping is also replace with `insn_t` with SSA-based control flow (by dominator graph traversal), and several opcodes are removed as being only used in `ph1_ir_t` but `insn_t`.

Notice that shecc still contains 2 phase IR transformation / generation after this patch. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request refactors the code by removing the deprecated phase 1 intermediate representation (`ph1_ir_t`) and replacing it with a more efficient SSA-based control flow. The changes enhance maintainability and performance by eliminating outdated opcodes and updating the IR dumping process. Overall, this modernizes the codebase for future enhancements.
<br>
<br>
<br>
<br>
</div>


DrXiao:
I'm unsure whether the [`Intermediate Representation`](https://github.com/sysprog21/shecc?tab=readme-ov-file#intermediate-representation) section in the `README.md` needs to be updated? It uses the `fib()` function and the corresponding IRs, which were previously produced by `ph1_ir_t`, to explain the meaning of IRs. Does the lastest IR output still match the previous one?

ChAoSUnItY:
>I'm unsure whether the [Intermediate Representation](https://github.com/sysprog21/shecc?tab=readme-ov-file#intermediate-representation) section in the README.md needs to be updated?

Yes, it would still need to be updated since now IR dumping delegates to SSA based control flow graph, which have some differences between previous one on label organization and control flow representation.  It would be a good chance to discuss how can we improve the SSA based control flow graph to look more "intuitive".

Here's the current 1 phase IR dumping generated by `insn_t` and `basic_block_t`:

```S
def int @fib(int %n) {
.bb.0:
        jmp .bb.2
.bb.2:
        const %.t0, 0
        %.t1 = eq %n, %.t0
        br %.t1, .bb.5, .bb.6
.bb.6:
        jmp .bb.9
.bb.9:
        const %.t3, 1
        %.t4 = eq %n, %.t3
        br %.t4, .bb.12, .bb.13
.bb.13:
        jmp .bb.16
.bb.16:
        const %.t6, 1
        %.t7 = sub %n, %.t6
        push %.t7
        call @fib
        retval %.t8
        const %.t9, 2
        %.t10 = sub %n, %.t9
        push %.t10
        call @fib
        retval %.t11
        %.t12 = add %.t8, %.t11
        ret %.t12
.bb.12:
        const %.t5, 1
        ret %.t5
.bb.5:
        const %.t2, 0
        ret %.t2
.bb.1:
}
```

fib
DrXiao, line 1: This file seems to be an executable generated by shecc, and I think it is unnecessary.

src/globals.c
```diff
@@ -43,17 +44,13 @@ hashmap_t *CONSTANTS_MAP;
...
-
-ph1_ir_t *PH1_IR;
-int ph1_ir_idx = 0;
-
 arena_t *INSN_ARENA;
 
 /* BB_ARENA is responsible for basic_block_t / ph2_ir_t allocation */
 arena_t *BB_ARENA;
 
+int dbg_bb_label_idx = 0;
```
jserv, line 52: Avoid using `dbg_` prefix. It is a matter of verbose messages instead of debugging purpose.

src/globals.c
```diff
@@ -1080,214 +1063,260 @@ void print_indent(int indent)
...
+                    continue;
+
+            print_indent(1);
+            printf("ret\n");
+        }
+
+        printf("}\n");
+    }
+
+    printf("==<END OF INSN DUMP>==\n");
```
jserv, line 1321: Where is the beginning?

Can you follow the way of Linux? See https://www.kernel.org/doc/html/v5.1/admin-guide/bug-hunting.html

jserv (changes requested):
Append `Close #197` at the end of git commit messages. Fix the grammar mistakes as well.

ChAoSUnItY:
The general purpose of #197 is not to remove phase 1 IR but to refactor `opcode_t` to respective `ph1_opcode_t` and `ph2_opcode_t`, I don't think this patch does the refactor.

src/globals.c
ChAoSUnItY: Should we also do the same thing in phase 2 IR?

jserv:
> The general purpose of #197 is not to remove phase 1 IR but to refactor `opcode_t` to respective `ph1_opcode_t` and `ph2_opcode_t`.

The subject of this pull request was misleading. You can proceed the removal of phase 1 IR right now though.

src/globals.c
jserv: > Should we also do the same thing in phase 2 IR?

Not necessary. We shall concentrate on the removal of phase 1 IR.

jserv (changes requested):
Rebase the latest `master` branch and resolve conflicts.

jserv:

PR #201, opened 2025-05-13
Stablize variable typing and support integer truncation and sign extension

### LINKED ISSUE #166: Certain operations lack consideration for data types
[ Edit: 2025/01/21 19:15 ]
Recently, I used the following tricky code to test shecc:
```c
/* test.c */
int main() {
    char a = 0x11, b = 0x22, c = 0x33, d = 0x44;
    char *aa = &a, *bb = &b, *cc = &c, *dd = &d;
    int *aaa = &a;
    char arr[4];
    arr[0] = 0x11;
    arr[1] = 0x22;
    arr[2] = 0x33;
    arr[3] = 0x44;

    /* 1 */
    printf("== 1 ==\n");
    printf("%d %d %d %d\n", a, b, c, d);
    printf("%x %x %x %x\n", aa, bb, cc, dd);
    printf("%d %d %d\n", bb - aa, cc - bb, dd - cc);
    printf("%x %x\n\n", *aaa, aaa[0]);

    /* 2 */
    aaa = arr;

    printf("== 2 ==\n");
    for (int i = 0; i < 4; i++)
	    printf("arr[%d] = %x\n", i, arr[i]);

    printf("aaa = %x, arr = %x\n", aaa, arr);
    printf("*aaa = %x, aaa[0] = %x\n\n", *aaa aaa[0]);
    
    /* 3 */
    printf("== 3 ==\n");
    a += 6000;
    b += 400;
    c -= 400;
    d -= 6000;
    aaa = &a;
    printf("a = %d, b = %d, c = %d, d = %d\n", a, b, c, d);

    return 0;
}
```
When the code is compiled using GCC with no optimization, all outputs match the expected results:
```
$ gcc -o test test.c
$ ./test
== 1 ==
17 34 51 68
1759c720 1759c721 1759c722 1759c723
1 1 1
44332211 44332211

== 2 ==
arr[0] = 11
arr[1] = 22
arr[2] = 33
arr[3] = 44
aaa = 1759c754, arr = 1759c754
*aaa = 44332211, aaa[0] = 44332211

== 3 ==
a = -127, b = -78, c = -93, d = -44
```
However, if using shecc to compile the code, some outputs seem to be wrong:
```
$ qemu-arm test
== 1 ==
17 34 51 68
407ffd8c 407ffd90 407ffd94 407ffd98
4 4 4
11 11

== 2 ==
arr[0] = 11
arr[1] = 22
arr[2] = 33
arr[3] = 44
aaa = 407ffda8, arr = 407ffda8
*aaa = 44332211, aaa[0] = 44332211

== 3 ==
a = 6017, b = 434, c = -349, d = -5932
```
By the code and the comment in `test.c`, it is obvious that the code is divided to 3 snippets. Next, I use these snippets to explain their behaviors and the bug what I found.

---
### snippet 1
After introducing the pull request (#171), **every stack allocation increment had been ensured to be properly aligned**. Due to stack alignment requirements, **each single character variable occupies 4 bytes**, despite a character's actual size being 1 byte.

Thus, the following results are correct.
1. the pointer arithmetic results like `bb - aa`, `cc - bb` and `dd - cc` are `4`.
2. Because `aaa` points to the address of `a`, the result of the dereference through `aaa` is equaivalent to the value of `a`.

---
### snippet 2
```diff
- printf("*aaa = %x, aaa[0] = %x\n\n", *arr, aaa[0]);    /* 2024/12/15 22:56 */
+ printf("*aaa = %x, aaa[0] = %x\n\n", *aaa, aaa[0]);    /* 2025/01/21 19:15 */
```
(Previously, I made a typo so that I misunderstood pointer dereferencing also had an error. After fixing it, the result is correct.)

---
### snippet 3 - **(bug)**
Because the variables (`a` ~ `d`) are signed characters, their results should be overflowed after executing additions and subtractions.

The correct results should be `a = -127`, `b = -78`, `c = -93` and  `d = -44`, but the results of the executable compiled by shecc produces error outputs.

## Conclusion
According to the result of the snippet 3, the current backends may lack consideration of the data type when generating load/store instructions.

For example, the Arm backend may generates a `ldr` instruction for loading a character (1 byte), regardless of the data type. But, For character variables, it must generate `ldrsb` instructions instead.

Since these problems also occur in the RISC-V backend, both the Arm and RISC-V backends require improvements to generate data-type-specific instructions, such as `ldrsb` for signed characters.

ChAoSUnItY:
This is a attempt to support variable typing for all local variables (including synthesized ones) so that integer truncation and sign extension could be implemented, which targets to fix #166.

Notice that there's still some works including type derivation from operators and code cleanup.

## About sign extension in rv32

It seems that package `qemu-user` is currently outdated (6.2.0) so Zbb extension is not supported, which results in the implementation of sign extension in rv32 currently using straightforward way to implement, that is bitwise shifting. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the variable typing system by implementing support for integer truncation and sign extension, optimizing memory allocation for local variables, and improving error handling. It addresses issue #166 and includes unit tests to ensure the reliability of these enhancements.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 2 - The changes are well-structured and documented, making the review process straightforward.
</div>


tests/driver.sh
DrXiao, line 1: Could you test the following case?
```c
int main()
{
    char a = 0xFF;
    int b = a;
    printf("%d %d\n", a, b);
    return 0;
}
```

When compiling the above code without the proposed changes, it prints `255 255`.

If I understand correctly, I think the proposed changes should handle this test case properly and produce  `-1 -1`, since these changes implement sign extension.

src/arm-codegen.c
```diff
@@ -421,6 +427,22 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
         emit(__mov_i(__NE, rd, 0));
         emit(__mov_i(__EQ, rd, 1));
         return;
+    case OP_trunc:
+        if (rm == 1) {
+            rm = 0xFF;
+        } else if (rm == 4) {
+            rm = 0xFFFFFFFF;
+        } else {
```
DrXiao, line 435: I think a shorter form can be used here:
```c
if (rm == 1)
    rm = 0xFF;
else if (rm == 4)
    rm = 0xFFFFFFFF;
else {
    ....
}
```

src/riscv-codegen.c
```diff
@@ -396,6 +402,24 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
         emit(__sltu(rd, __zero, rs1));
         emit(__xori(rd, rd, 1));
         return;
+    case OP_trunc:
+        if (ph2_ir->src1 == 1) {
+            rs2 = 0xFF;
+        } else if (ph2_ir->src1 == 4) {
+            rs2 = 0xFFFFFFFF;
+        } else {
```
DrXiao, line 410: Ditto.

tests/driver.sh
ChAoSUnItY: Yes, I missed lots of assignment resizing, apparently this means that we would need to introduce a general function that handles both `var_t` and `lvalue_t` that resizes source `var_t` to appropriate size by either `OP_trunc` or `OP_sign_ext`.

src/arm-codegen.c
```diff
@@ -421,6 +427,22 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
...
+            printf("Unsupported truncation operation with target size %d\n",
```
jserv, line 436: Use `error` when possible.

src/arm-codegen.c
```diff
@@ -421,6 +427,22 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
...
+            abort();
+        }
+
+        emit(__and_i(__AL, rd, rn, rm));
+        return;
+    case OP_sign_ext:
+        emit(__sxtb(__AL, rd, rn, 0));
+        return;
     default:
         printf("Unknown opcode\n");
```
jserv, line 447: Ditto.

src/arm.c
```diff
@@ -349,3 +354,14 @@ int __teq(arm_reg rd)
 {
     return __mov(__AL, 1, arm_teq, 1, rd, 0, 0);
 }
+
+int __sxtb(arm_cond_t cond, arm_reg rd, arm_reg rm, int rotation)
+{
+    if (rotation != 0 && rotation != 8 && rotation != 16 && rotation != 24) {
+        printf("SXTB rotation must be 0, 8, 16, or 24\n");
+        abort();
```
jserv, line 362: Use `error`.

src/arm-codegen.c
ChAoSUnItY: The issue of using error here is that it would leads to incorrect error diagnostic location, which is kind of meaningless?

ChAoSUnItY: I would like to open a new issue regards this issue since `error` is not capable of passing variadic arguments and it would make error message being limited.

ChAoSUnItY:
Regards to the issue of sign extension on rv32 backend, should we introduce a compiler parameter `-Zbb` to allow user to choose whether generate sign extension instruction with `sext.b` (or `sext.h` in the future once short type is implemented) or with bitwise shifts?

src/arm-codegen.c
jserv: > I would like to open a new issue regards this issue since `error` is not capable of passing variadic arguments and it would make error message being limited.

Go ahead.

jserv:
> Regards to the issue of sign extension on rv32 backend, should we introduce a compiler parameter `-Zbb` to allow user to choose whether generate sign extension instruction with `sext.b` (or `sext.h` in the future once short type is implemented) or with bitwise shifts?

It is a good idea. However, we then need to an effective way to select the extensions for RISC-V backend.

src/globals.c
```diff
@@ -1018,6 +1016,14 @@ void global_release()
     strbuf_free(elf_section);
 }
 
+/* Reports an error without specifying a position */
+void error_no_loc(char *msg)
```
jserv, line 1020: Let's rename it to `fatal`. Such helper functions should be always having shorter names.

src/globals.c
```diff
@@ -1286,11 +1302,11 @@ void dump_insn()
         for (int i = 0; i < func->num_params; i++) {
             if (i != 0)
                 printf(", ");
-            printf("%s", func->param_defs[i].type_name);
+            printf("%s", func->param_defs[i].type->type_name);
 
             for (int k = 0; k < func->param_defs[i].is_ptr; k++)
                 printf("*");
-            printf(" %%%s", func->param_defs[i].var_name);
+            printf(" %%%s", func->param_defs[i].type->type_name);
```
DrXiao, line 1309: Is this change necessary? The original code prints `var_name` instead of `type_name`.

src/parser.c
```diff
@@ -38,12 +38,67 @@ char *gen_name_to(char *buf)
...
 {
-    if (blk->next_local >= MAX_LOCALS)
-        error("Too many locals");
+    var_list_t *var_list = &blk->locals;
 
-    var_t *var = &blk->locals[blk->next_local++];
+    if (var_list->size >= var_list->capacity) {
+        var_list->capacity *= 2;
+
+        var_t **new_locals = calloc(var_list->capacity, sizeof(var_t *));
```
DrXiao, line 46: In `add_block()`, `locals.elements` receives the pointer returned by `arena_alloc()`, so I am confused why `calloc()` is used here.

Would it be better to use `arena_alloc()` here as well, for consistency?

src/globals.c
ChAoSUnItY: It's a typo while I was removing `type_name` field from `var_t`, will change it back in next commit.

src/parser.c
```diff
@@ -167,6 +223,73 @@ opcode_t get_operator()
     return op;
 }
 
+var_t *promote_unchecked(block_t *block,
```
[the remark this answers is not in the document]
ChAoSUnItY: Shorten naming is good in my opinion, but my thought on `__` prefix is that functions using this naming should be some kind of private functions that should not be used directly (see `lib/c.c`)? 

Fyi `_unchecked` suffix is inspired by Rust's naming, e.g. [`pointer#get_unchecked`](https://doc.rust-lang.org/std/primitive.pointer.html#method.get_unchecked_mut).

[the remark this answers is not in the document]
DrXiao: I understand your reasoning and don't have a better idea for shortening the name, so feel free to ignore this suggestion.

DrXiao (approved):

vacantron (approved):
LGTM

jserv:

PR #203, opened 2025-05-18
Refine the build system

### LINKED ISSUE #44: Validate the prerequisites when "make config" is executed
The build target "config" should not only specify the target for code generation (Arm by default) but also check the availability of prerequisites such as `qemu-arm` and `qemu-riscv32`.

DrXiao:
This pull request resolves the following issues:

### Fix empty `TARGET_EXEC` variable
Previously, I noticed that the build system directly executed the stage1 compiler to build the stage2 compiler, instead of invoking `qemu-user`. This issue could be observed by running `make VERBOSE=1`
```diff
$ make VERBOSE=1
...
...
+out/shecc-stage1.elf -o out/shecc-stage2.elf src/main.c
...
```
That is, `TARGET_EXEC`, which stores the the name of an architecture-specific qemu executable, is left empty and not properly assigned.

Therefore, these changes fix it to ensure the build system invokes qemu as expected.
 
Close #44 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the build system by fixing the empty TARGET_EXEC variable issue, which caused build failures. It improves Makefiles for ARM and RISC-V architectures with prerequisite checks for necessary tools, ensuring a more reliable build process and correct invocation of qemu executables.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,6 @@ CFLAGS := -O -g \
 BUILD_SESSION := .session.mk
 
 include mk/common.mk
-include mk/arm.mk
-include mk/riscv.mk
 -include $(BUILD_SESSION)
 
 STAGE0 := shecc
@@ -43,11 +41,7 @@ all: config bootstrap
 ifeq (,$(filter $(ARCH),$(ARCHS)))
 $(error Support ARM and RISC-V only. Select the target with "ARCH=arm" or "ARCH=riscv")
 endif
-
-ifneq ("$(wildcard $(PWD)/config)","")
-TARGET_EXEC := $($(shell head -1 config | sed 's/.*: \([^ ]*\).*/\1/')_EXEC)
-endif
-export TARGET_EXEC
+include mk/$(ARCH).mk
 
 config:
 	$(Q)ln -s $(PWD)/$(SRCDIR)/$(ARCH)-codegen.c $(SRCDIR)/codegen.c
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ CFLAGS := -O -g \
 
 BUILD_SESSION := .session.mk
 
-include mk/common.mk
 -include $(BUILD_SESSION)
 
 STAGE0 := shecc
@@ -26,6 +25,7 @@ STAGE2 := shecc-stage2.elf
 OUT ?= out
 ARCHS = arm riscv
 ARCH ?= $(firstword $(ARCHS))
+HOST_ARCH = $(shell arch 2>/dev/null)
 SRCDIR := $(shell find src -type d)
 LIBDIR := $(shell find lib -type d)
 
@@ -42,12 +42,14 @@ ifeq (,$(filter $(ARCH),$(ARCHS)))
 $(error Support ARM and RISC-V only. Select the target with "ARCH=arm" or "ARCH=riscv")
 endif
 include mk/$(ARCH).mk
+include mk/common.mk
 
 config:
 	$(Q)ln -s $(PWD)/$(SRCDIR)/$(ARCH)-codegen.c $(SRCDIR)/codegen.c
-	$(call $(ARCH)-specific-defs) > $@
+	$(Q)$(PRINTF) $(ARCH_DEFS) > $@
 	$(VECHO) "Target machine code switch to %s\n" $(ARCH)
 	$(Q)$(MAKE) $(BUILD_SESSION) --silent
+	$(Q)$(CONFIG_CHECK_CMD)
 
 $(OUT)/tests/%.elf: tests/%.c $(OUT)/$(STAGE0)
 	$(VECHO) "  SHECC\t$@\n"
@@ -103,6 +105,7 @@ $(OUT)/$(STAGE0): $(OUT)/libc.inc $(OBJS)
 	$(Q)$(CC) $(OBJS) -o $@
 
 $(OUT)/$(STAGE1): $(OUT)/$(STAGE0)
+	$(Q)$(STAGE1_CHECK_CMD)
 	$(VECHO) "  SHECC\t$@\n"
 	$(Q)$(OUT)/$(STAGE0) --dump-ir -o $@ $(SRCDIR)/main.c > $(OUT)/shecc-stage1.log
 	$(Q)chmod a+x $@
--- a/mk/common.mk
+++ b/mk/common.mk
@@ -5,8 +5,6 @@ else
     PRINTF = env printf
 endif
 
-HOST_ARCH = $(shell arch 2>/dev/null)
-
 # Control the build verbosity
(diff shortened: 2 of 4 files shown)


mk/check.mk
```diff
@@ -0,0 +1,10 @@
+# Check the prerequisites
+PREREQ_LIST := dot jq $(ARCH_EXEC)
+PREREQ_EXEC := $(shell which $(PREREQ_LIST))
+PREREQ_MISSING := $(filter-out $(notdir $(PREREQ_EXEC)),$(PREREQ_LIST))
+
+ifdef PREREQ_MISSING
+$(warning "Necessary packages: $(PREREQ_LIST)")
+$(warning "Missing packages: $(PREREQ_MISSING)")
+$(error "Please check package installation")
+endif
```
jserv, line 10: Move the content back to `mk/common.mk`

jserv:
If target architecture does not match host (e.g., configuring Arm32 on x86-64), the build system should warn and stop at stage 0.

DrXiao:
> If target architecture does not match host (e.g., configuring Arm32 on x86-64), the build system should warn and stop at stage 0.

I am confused by this description and not sure if I understood correctly. If we configure Arm32 on x86-64, the build system should proceed with bootstrapping. The stage 2 compiler would be built by the stage 1 compiler running through `qemu-arm`, rather than stopping the build

jserv:
> If we configure Arm32 on x86-64, the build system should proceed with bootstrapping.

Yes, it is valid when qemu-arm or qemu-riscv32 emulators are properly configured in the environment. However, the current implementation lacks comprehensive documentation or messages that would explain the emulation process, expected outputs, and potential limitations. Adding verbose explanations would significantly improve usability and troubleshooting capabilities for developers working across different architectures.

DrXiao:
For the current build system, consider the following table, which lists the valid host-target combinations for building shecc:

| host architecture | target architecture for shecc | Remark |
|--------------------------|-------------------------------------------| ------------|
| x86-64                   | armv7                        | use `qemu-arm` to emulate |
| x86-64                   | riscv                           | use `qemu-riscv` to emulate |
| armv7                    | armv7                       | run directly on the host computer|

> Adding verbose explanations would significantly improve usability and troubleshooting capabilities for developers working across different architectures.

If I understood correctly, for example:
* Use a computer with an `armv7` architecture to build shecc targeting riscv.
* Use a computer with a `powerpc` architecture to build shecc targeting armv7 or riscv.

The above builds should be considered invalid because these host-target combinations are not listed in the table.

That is, in the first case, we do not allow developers to build shecc targeting riscv on an ARM machine. In the second case, we also do not allow developers build shecc targeting arm or riscv on a non-x86-64 machine.

Therefore, in such cases, the build system should stop and output a warning message to inform the developer.

Is my understanding correct?

jserv:
> Is my understanding correct?

Yes, all architectures/platforms are capable of running stage0 once an ordinary C compiler is found. There is explicitly a notable prerequisite to enter stage1 and eventually the following stage2.

DrXiao:
I have adjusted the original way to perform the validation. The build system will check the required packages and print the warning messages when `make config`. However, if qemu is required but missing, the build will proceed up to stage 0, print warning messages, and stop at the beginning of stage 1.

I tested my the behavior using my beaglebone black (running Debian OS) by running `make ARCH=riscv`, and the build system behaved as expected:
```shell
debian@BeagleBone:~/shecc$ make distclean && make ARCH=riscv
env printf "ARCH=riscv" > .session.mk
Target machine code switch to riscv
Warning: missing packages: dot jq qemu-riscv32
Warning: Please check package installation
  CC+LD out/inliner
  GEN   out/libc.inc
  CC    out/src/main.o
  LD    out/shecc
Warning: failed to build the stage 1 and stage 2 compilers due to missing qemu-riscv32
make: *** [Makefile:108: out/shecc-stage1.elf] Error 1
```
If running `make` again, the build system simply shows the warning.
```shell
debian@BeagleBone:~/shecc$ make
Warning: failed to build the stage 1 and stage 2 compilers due to missing qemu-riscv32
make: *** [Makefile:108: out/shecc-stage1.elf] Error 1
```

jserv (changes requested):
Both `mk/arm.mk` and `mk/riscv.mk` have common rules, and it is better to use `define` to unify them.

Makefile
```diff
@@ -43,17 +41,15 @@ all: config bootstrap
...
-ifneq ("$(wildcard $(PWD)/config)","")
-TARGET_EXEC := $($(shell head -1 config | sed 's/.*: \([^ ]*\).*/\1/')_EXEC)
-endif
-export TARGET_EXEC
+include mk/$(ARCH).mk
+include mk/common.mk
 
 config:
 	$(Q)ln -s $(PWD)/$(SRCDIR)/$(ARCH)-codegen.c $(SRCDIR)/codegen.c
 	$(call $(ARCH)-specific-defs) > $@
```
jserv, line 49: Add `$(Q)` prior to call.

.github/workflows/main.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         githubToken: ${{ github.token }}
         install: |
           apt-get update -qq -y
-          apt-get install -yqq build-essential
+          apt-get install -yqq build-essential graphviz jq
```
DrXiao, line 42: This change may become unnecessary, and I will fix them.

mk/common.mk
```diff
@@ -23,3 +21,29 @@ PASS_COLOR = \e[32;01m
...
+	PREREQ_LIST += $(ARCH_QEMU)
+	ifeq ($(filter $(ARCH_QEMU),$(notdir $(shell which $(ARCH_QEMU)))),)
+		STAGE1_WARN_MSG := "Warning: failed to build the stage 1 and $\
+						   	stage 2 compilers due to missing $(ARCH_QEMU)\n"
+		STAGE1_CHECK_CMD := $(VECHO) $(STAGE1_WARN_MSG) && exit 1
+	endif
+
+	# Generate the path of an architecture-specific qemu
+	TARGET_EXEC = $(shell which $(ARCH_QEMU))
+endif
```
jserv, line 39: Use 4 spaces for indention.

mk/common.mk
```diff
@@ -23,3 +21,29 @@ PASS_COLOR = \e[32;01m
...
+export TARGET_EXEC
+
+PREREQ_EXEC := $(shell which $(PREREQ_LIST))
+PREREQ_MISSING := $(filter-out $(notdir $(PREREQ_EXEC)),$(PREREQ_LIST))
+
+ifdef PREREQ_MISSING
```
jserv, line 45: Ditto. Be aware of the consistent indention.

jserv:

PR #204, opened 2025-05-18
Allow parameter list to be a single "void"

lumynou5:
A parameter list is either a single keyword "void" or a comma-separated list of parameters according to the standard.  Previously, only the latter is supported, and the former leads to unexpected token errors. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the parser's parameter list handling by allowing a single 'void' keyword as a valid parameter list, in addition to the existing comma-separated format. This change improves flexibility, resolves previous unexpected token errors, and standardizes function signatures across multiple files, contributing to cleaner code. It also includes new tests to ensure proper handling of parameter declarations, improving parser robustness.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -108,7 +108,7 @@ void opstack_push(var_t *var)
     operand_stack[operand_stack_idx++] = var;
 }
 
-var_t *opstack_pop()
+var_t *opstack_pop(void)
 {
     return operand_stack[--operand_stack_idx];
 }
@@ -179,7 +179,7 @@ int get_unary_operator_prio(opcode_t op)
     }
 }
 
-opcode_t get_operator()
+opcode_t get_operator(void)
 {
     opcode_t op = OP_generic;
     if (lex_accept(T_plus))
@@ -317,7 +317,7 @@ int read_numeric_constant(char buffer[])
     return value;
 }
 
-int read_constant_expr_operand()
+int read_constant_expr_operand(void)
 {
     char buffer[MAX_ID_LEN];
     int value;
@@ -449,7 +449,7 @@ int read_constant_infix_expr(int precedence)
     return lhs;
 }
 
-int read_constant_expr()
+int read_constant_expr(void)
 {
     return read_constant_infix_expr(0);
 }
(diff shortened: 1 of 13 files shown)


src/parser.c
```diff
@@ -596,6 +596,17 @@ void read_parameter_list_decl(func_t *func, int anon)
 {
     int vn = 0;
     lex_expect(T_open_bracket);
+
+    char token[MAX_TYPE_LEN];
+    if (lex_peek(T_identifier, token) && !strncmp(token, "void", 4)) {
+        next_token = lex_token();
+        if (lex_accept(T_close_bracket))
```
ChAoSUnItY, line 603: As [C99 specification](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2310.pdf) specifies:

> 6.7.6.3 Function declarators (including prototypes)
> ...
> 10 The special case of an unnamed parameter of type void as the only item in the list specifies that the
function has no parameters.
> ...

This indicates that any forms other than `f(void)` is incorrect, e.g.
```c
f(void n)
f(void, int n)
f(void n, int n)
```

That is, you should expect `T_close_bracket` and rather than any other parameter list identifiers.

ChAoSUnItY (changes requested):
Please add tests in `tests/driver.sh` to validate your work follows specification's description.

DrXiao:
The current shecc also contains functions such as `global_init()` and `global_release()` that have no parameters., I think the proposed changes can modify them as well by explicitly adding `void` to their parameter lists. Therefore, the bootstrapping process can also validate whether the usage of `void` is correctly parsed.

src/parser.c
```diff
@@ -596,6 +596,21 @@ void read_parameter_list_decl(func_t *func, int anon)
...
+            return;
+        strcpy(func->param_defs[vn].type_name, token);
```
ChAoSUnItY, line 605: Latest change (#201) deprecated `var_t#type_name`. Instead, you should use `read_full_var_decl` to properly initialize the parameter `var_t` then check if `func->param_defs[vn].type == TY_void` to assert later criteria.

ChAoSUnItY (changes requested):
Rebase the latest `master` branch and resolve the error.

src/parser.c
lumynou5: `read_full_var_decl` cannot handle both cases of named and unnamed.  If `anon` is zero, it expects an identifier and abort on void parameter lists; if `anon` is non-zero, it doesn't correctly read the first parameter with types such as `void *`.  The original implementation is actually a custom version of `read_partial_var_decl`.

Calling the function twice with different `anon` set can handle legal cases, but will allow things like `void [] name []`.  So, the solution seems same but with a newer way to construct `var_t`.

jserv:
Thank @lumynou5 for contributing!

PR #207, opened 2025-05-21
Support global string initialization

ChAoSUnItY:
This patch supports global string initialization, and allows integer values to be used as address value of any pointer type. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances global string initialization by allowing integer values as address values for pointer types. Key modifications include parser updates for string literals, improved register allocation for data addresses, and SSA updates for better string formatting. New tests have been added to ensure the functionality works as intended.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/tests/driver.sh
+++ b/tests/driver.sh
@@ -1357,6 +1357,19 @@ int main()
 }
 EOF
 
+# global string initialization and modification
+try_output 0 "Hello World!Hallo World!" << EOF
+char *data = "Hello World!";
+
+int main(void)
+{
+    printf(data);
+    data[1] = 'a';
+    printf(data);
+    return 0;
+}
+EOF
+
 # global initialization with logical and equality operation
 try_ 4 << EOF
 int b1 = 1 && 1;
--- a/src/parser.c
+++ b/src/parser.c
@@ -2078,10 +2078,19 @@ bool read_global_assignment(char *token)
 {
     var_t *vd, *rs1, *var;
     block_t *parent = GLOBAL_BLOCK;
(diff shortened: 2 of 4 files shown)


jserv:

tests/driver.sh
```diff
@@ -1357,6 +1357,19 @@ int main()
...
 EOF
 
+# global string initialization and modification
+try_output 0 "Hello World!Hallo World!" << EOF
+char *data = "Hello World!";
+
+int main(void)
+{
+    printf(data);
+    data[1] = 'a';
```
DrXiao, line 1367: I think this write operation is invalid because the string literal (`"Hello World!"`) should be read-only and placed in the `.rodata` section, although the current implementation does not create this section.

ChAoSUnItY: According to C99 `6.4.5 String Literals`:
> 7. It is unspecified whether these arrays are distinct provided their elements have the appropriate
values. If the program attempts to modify such an array, the behavior is undefined.

The common practice in other well-known compiler (gcc, clang) suggest that this is indeed an invalid operation  since this would cause segmentation fault.

The reason I submitted this initial support patch is to allow my local work to declare string litearl in global scope, which is kind of rough but working anyway. Would you like to look into this behavior patch?

DrXiao: Understood. I think we can create a patch to add `TODO` or `FIXME` comment to inform developers that this write behavior should be fixed and avoided in the future.

PR #217, opened 2025-06-12
Fix incorrect usage of printf

ChAoSUnItY:
This fixes potential invalid pointer dereference when fails to allocate.  
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request fixes a bug in the `src/globals.c` file by correcting the usage of the `printf` function. The update ensures that the error message accurately reflects the size of the allocation during memory allocation failures, mitigating potential invalid pointer dereference issues.

</div>


jserv:

PR #218, opened 2025-06-12
Enhance workflow procedure

ChAoSUnItY:
This patch enhances workflow procedure by separating each critical checks from single step to multiple steps, allows developers and reviewers to better locate which check fails instead of checking raw output log.


jserv:

PR #219, opened 2025-07-13
Fix parsing to support uppercase 0X hex literals

fourcolor:
Previously, literals with uppercase 0X prefix (e.g., "0XABC") were misparsed as invalid. This commit fixes that and improves compatibility with C99-style numeric constants.

- Removed incorrect inclusion of 'x' as a hex digit in is_hex()
- Updated is_numeric() to skip 0x/0X prefix when validating hex digits
- Adjusted read_numeric_constant() to consistently handle both 0x and 0X prefixes 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances numeric literal parsing by adding support for uppercase 0X hexadecimal prefixes, correcting previous misparsing issues. Modifications were made to the is_hex and is_numeric functions for better validation, and the read_numeric_constant function was updated for consistent handling of both prefix styles, improving C99 compatibility. New tests have been included to ensure the accuracy of these changes.

</div>

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -42,8 +42,8 @@ bool is_digit(char c)
 
 bool is_hex(char c)
 {
-    return ((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || c == 'x' ||
-            (c >= 'A' && c <= 'F'));
+    return (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') ||
+           (c >= 'A' && c <= 'F');
 }
 
 bool is_numeric(char buffer[])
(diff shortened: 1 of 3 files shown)


jserv (changes requested):
Run `clang-format -i` prior to submit.

jserv (changes requested):
Consolidate the test suite to reflect the proposed changes.

src/lexer.c
```diff
@@ -42,19 +42,19 @@ bool is_digit(char c)
 
 bool is_hex(char c)
 {
-    return ((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || c == 'x' ||
```
DrXiao, line 45: How about adding `c == 'X'` to the expression instead of removing `c == 'x'` ?

If so, the proposed changes could be refined accordingly.

fourcolor: I think we should avoid treating 'x' or 'X' as valid hex digits, since they are part of the prefix (0x or 0X) but not valid hex characters themselves.

jserv:
Thank @fourcolor for contributing!

PR #220, opened 2025-07-15
Fix potential segfault when reporting error

ChAoSUnItY:
When reaching to SSA phase, the `error` function would always cause segment fault due to the internal design, which is the source index is always going to be out of bound. In this patch, we instead use `fatal` function to avoid this issue. 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request fixes a critical bug in the SSA phase by replacing the 'error' function with 'fatal', preventing segmentation faults from out-of-bounds indices. This change significantly enhances the stability of the code during execution.

</div>

### DIFF
--- a/src/ssa.c
+++ b/src/ssa.c
@@ -697,7 +697,7 @@ void new_name(block_t *block, var_t **var)
 var_t *get_stack_top_subscript_var(var_t *var)
 {
     if (var->base->rename.stack_idx < 1)
-        error("Index is less than 1");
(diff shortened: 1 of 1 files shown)


jserv:
> When reaching to SSA phase, the `error` function would always cause segment fault due to the internal design, which is the source index is always going to be out of bound. In this patch, we instead use `fatal` function to avoid this issue.

Create a dedicated issue to raise the above.

jserv:

PR #225, opened 2025-08-02
Add va_list typedef and variadic function support

### LINKED ISSUE #202: Implement `va_list` and related functions
Currently in shecc, variadic functions is lack of flexibility, since variadic arguments cannot be passed from a variadic function to anthoer directly. This means we'll need a shecc-defined `va_list` and functions like `vprintf` in order to improve current error diagnostic mechanism in shecc.

However, due to current shecc's frontend, we may need to find a way to properly address issues of macro definitions such as `va_start` and `va_end`.

jserv:
This commit adds va_list typedef for basic variadic function support and implement typedef pointer handling in parser (typedef char *string).

Close #202


src/parser.c
```diff
@@ -2956,6 +2961,14 @@ void read_global_statement(void)
             type->base_type = base->base_type;
             type->size = base->size;
             type->num_fields = 0;
+            type->is_ptr = 0;
+
+            /* Handle pointer types in typedef: typedef char *string; */
+            while (lex_accept(T_asterisk)) {
+                type->is_ptr++;
+                type->size = 4; /* pointer size is 4 bytes */
```
DrXiao: How about using `PTR_SIZE` to set `type->size`?

tests/driver.sh
DrXiao, line 1: Because the proposed changes also enable shecc to support  handling of typedef pointers, how about adding some test cases to verify its functionality?

e.g.:
```c
typedef char *string;
int main(void)
{
    string str = malloc(sizeof(char) * 100);
    strcpy(str, "Hello world");
    printf("%s\n", str);
    free(str);
    return 0;
}
```

src/parser.c
```diff
@@ -657,7 +657,9 @@ void read_parameter_list_decl(func_t *func, int anon);
 void read_inner_var_decl(var_t *vd, int anon, int is_param)
 {
     vd->init_val = 0;
-    vd->is_ptr = 0;
+    /* Preserve typedef pointer level - don't reset if already inherited */
+    int base_ptr_level = vd->is_ptr;
+    vd->is_ptr = base_ptr_level;
```
DrXiao: Does `base_ptr_level` serve any other purpose? I am not sure why this variable is necessary.

Issue #205, opened 2025-05-18
Thoughts on memory usage reduction


DrXiao:
When submitting the pull request (#203 ), I used my [beaglebone black](https://www.beagleboard.org/boards/beaglebone-black) (running Debian OS) to build shecc targeting armv7.

 However, I found that the current memory usage is too high to build the stage 2 compiler on my beaglebone black.
```shell
debian@BeagleBone:~/shecc$ make
env printf "ARCH=arm" > .session.mk
Target machine code switch to arm
Warning: missing packages: dot jq
Warning: Please check package installation
  CC+LD out/inliner
  GEN   out/libc.inc
  CC    out/src/main.o
  LD    out/shecc
  SHECC out/shecc-stage1.elf
  SHECC out/shecc-stage2.elf
[ 1830.930466] Out of memory: Killed process 1995 (shecc-stage1.el) total-vm:1875560kB, anon-rss:429492kB, file-rss:148kB, shmem-rss:0kB, UID:1000 pgtables:1832kB oom_score_adj:0
make: *** [Makefile:115: out/shecc-stage2.elf] Killed
```

It would be better to reduce the memory usage so that the build process can complete even on low-memory ARM machines.

jserv:
@DrXiao, you can adapt the above PoC implementation and see if it works on Beaglebone hardware (without swap).

DrXiao:
The improvement is still a work in progress because I have encountered a troublesome problem.

After enhancing the `malloc`/`free` routines, the bootstrap process fails with **an unknown-cause segfault** when building the stage 2 compiler.

jserv:
@DrXiao, please see if #226 works for Debian GNU/Linux running [beaglebone black](https://www.beagleboard.org/boards/beaglebone-black).

DrXiao:
I tried building shecc on beaglebone black again.

Initially, I switched to commit 8f2b234 to build shecc, and the bootstrapping process failed as expected. However, I noticed that when switching to the latest commit on the master branch (1fb9fa5) , the bootstrapping completed successfully on my board.

Therefore, I used `git-bisect` to identify the commit that implicitly resolved the memory issue, and I found the bootstrapping started to succeed after commt 4be720b was merged to the master branch.

Next, I will try to analyze the memory usage of 8f2b234, 1fb9fa5 4be720b and  #226 .

DrXiao:
> Next, I will try to analyze the memory usage of 8f2b234, 1fb9fa5 4be720b and #226 .

[...]

PR #226, opened 2025-08-02
Harden malloc implementation

jserv:
The original malloc was development-only with critical flaws:
- No alignment
- Overflow vulnerability (calloc multiplication)
- Null pointer crashes (cleanup functions)
- Complex allocation logic
    
New implementation adds:
- 8-byte alignment for cache optimization
- Bounds checking preventing integer overflow
- Null-safe memory management operations 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the malloc implementation by fixing alignment issues, preventing overflow vulnerabilities, and addressing null pointer crashes. It introduces 8-byte alignment for cache optimization, bounds checking for integer overflow, and null-safe memory management, improving reliability and performance.

</div>


DrXiao:
I fetched the updated commits on my beaglebone black and rebuilt shecc. However, when using 3802292, the build process encountered a segfault issue during the stage 2 compiler build.

```shell
debian@BeagleBone:~/shecc$ make
env printf "ARCH=arm" > .session.mk
Target machine code switch to arm
Warning: missing packages: dot jq
Warning: Please check package installation
  CC+LD out/inliner
  GEN   out/libc.inc
  CC    out/src/main.o
  LD    out/shecc
  SHECC out/shecc-stage1.elf
  SHECC out/shecc-stage2.elf
make: *** [Makefile:129: out/shecc-stage2.elf] Segmentation fault
```

jserv:
> I fetched the updated commits on my beaglebone black and rebuilt shecc. However, when using [3802292](https://github.com/sysprog21/shecc/commit/3802292dc1bab93fe5a8f8cdb96f0ba5607d1a09), the build process encountered a segfault issue during the stage 2 compiler build.

See if commit dc66c251d8db908af49b83bfb07680008a36aabd helps.

jserv:
For further improvements, it depends on #230 for richer expressions.

DrXiao:
> See if commit dc66c25 helps.

After fetching this commit, the build still fails when compiling the stage 2 compiler on my beaglebone board.

DrXiao:
Since several pull requests have been merged recently, it may be necessary to rebase the latest `master` branch again for the build process to succeed on the beaglebone board.

If the build still fails on the beaglebone black after rebasing, we should review the proposed changes or the previously merged commits to identify any potential issues.

DrXiao:
I have fetched the latest `master` branch and build shecc on my board again. Unexpectedly, the build also fails due to a segfault. Next, I will use `git-bisect` to find the root cause.

jserv:
> I have fetched the latest `master` branch and build shecc on my board again. Unexpectedly, the build also fails due to a segfault. Next, I will use `git-bisect` to find the root cause.

@ChAoSUnItY, can you confirm with Raspberry Pi + 32-bit Raspbian/Debian?

ChAoSUnItY:
> can you confirm with Raspberry Pi + 32-bit Raspbian/Debian?

I've confirmed the following context:

1. It indeed crashes when building stage 1 compiler.
2. It crashes when parsing the parameter `shift_type shift` in declaration of function `__srl_amt` defined in `arm.c`.
3. When parsing it, the type is found and it returns a nonnull pointer, but GDB indicates that the pointer it returns is not accessable (this occurs on line 721), which triggers segment fault.
https://github.com/sysprog21/shecc/blob/dc66c251d8db908af49b83bfb07680008a36aabd/src/parser.c#L704-L724

Judging from above information, I suspect this could be memory corruption done by incorrect memory handing.

jserv:
> Judging from above information, I suspect this could be memory corruption done by incorrect memory handing.

Check #237

DrXiao:
As shown in https://github.com/sysprog21/shecc/issues/205#issuecomment-3164314390 , the hybrid allocation strategy seems to have no significant impact on reducing memory use.

After reviewing the current source code of shecc, I noticed that most of `malloc()` calls are used to allocate large objects, so the proposed changes don't significantly improve the memory efficiency, since small objects are already allocated space using hashmap or the arena allocator.

PR #230, opened 2025-08-05
Support complex pointer arithmetic in dereference

jserv:
Previously, the parser only handled simple identifier dereference (*var) and would crash on complex expressions like *(ptr + offset). This change extends the dereference operator handling to accept general expressions in parentheses.

This enables compilation of previously failing patterns:
- *(p + 4) - direct offset
- *(p + i + 2) - variable in expression
- *(p + i * 2) - arithmetic in expression

It also handles consecutive asterisks ('**pp', '***ppp') by counting dereference levels and applying them iteratively.


ChAoSUnItY (commented):
Logic changes look fine, but maybe adding some additional test suites to show that complex pointer arithmetics on both LHS and RHS are correctly parsed would be better?

PR #232, opened 2025-08-05
Add support for testing large return values (> 255)

### LINKED ISSUE #152: Improve test suite for a broader range of cases more efficiently
As noted in #151 , the current test suite only accepts tests where the final result is between 0 and 255 (a byte). This restriction makes it difficult to perform certain tests.

For example, when testing multi-digit hexadecimal values with the Fibonacci function, such as `fib(0x10) = 987`, the test must be written as:

```c
(fib(0x10) - 987) == 0
```
Because final result exceed 255. This approach is somewhat awkward and unintuitive. Ideally, we should be able to test the result directly without needing to fit everything within the byte range. Expanding the range of accepted results or modifying the mechanism for capturing return values could make the test suite more flexible and easier to use for larger values.

jserv:
Previously, test suite was limited to verifying return values between 0-255 due to Unix exit code constraints. It made testing functions with larger return values awkward, requiring workarounds like "(fib(16) - 987) == 0".

This commit introduces a new try_large() function that:
- Wraps test code in a test_function() that returns the value to test
- Uses printf to output the return value instead of exit codes
- Compares the printed output against expected values

Close #152


fennecJ:
Looks good to me. This approach using printf makes it easier for future contributors to extend support to non-integer types if needed. It also allows us to test the printf formatting of large integers (e.g., %x, %o).

PR #235, opened 2025-08-06
Replace strcmp chains with hashmap lookups

jserv:
This commit modernizes lexer token recognition by utilizing existing hashmap infrastructure and arena allocation system for performance and consistency, resulting in O(n) → O(1) average case token lookup.

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -1113,8 +1113,14 @@ void global_init(void)
     elf_section = strbuf_create(MAX_SECTION);
 }
 
+/* Forward declaration for lexer cleanup */
+void lexer_cleanup(void);
+
 void global_release(void)
 {
+    /* Cleanup lexer hashmaps */
+    lexer_cleanup();
+
     hashmap_free(MACROS_MAP);
     free(TYPES);
     arena_free(BLOCK_ARENA);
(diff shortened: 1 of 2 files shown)


jserv:
Due to internal error of parser, I can not write elegant code like below:
```c
// Create preprocessor directive lookup table
typedef struct {
    const char *name;
    token_t token;
} directive_entry_t;

static const directive_entry_t directives[] = {
    {"#include", T_cppd_include}, {"#define", T_cppd_define},
    {"#undef", T_cppd_undef}, {"#error", T_cppd_error},
    {"#if", T_cppd_if}, {"#elif", T_cppd_elif},
    {"#ifdef", T_cppd_ifdef}, {"#ifndef", T_cppd_ifndef},
    {"#else", T_cppd_else}, {"#endif", T_cppd_endif},
    {"#pragma", T_cppd_pragma}, {NULL, T_start}
};

// Use binary search instead of linear strcmp chain
token_t lookup_directive(const char *token) {
    int low = 0, high = 10;  // Number of directives - 1
    while (low <= high) {
        int mid = (low + high) / 2;
        int cmp = strcmp(token, directives[mid].name);
        if (cmp == 0) return directives[mid].token;
        else if (cmp < 0) high = mid - 1;
        else low = mid + 1;
    }
    return T_identifier;
}
```

Instead, I have to specify each item for hash map.

PR #237, opened 2025-08-07
Prevent buffer overflow in type array allocation

jserv:
This commit adds bounds checking to add_type and increase MAX_TYPES to 128, fixing memory corruption during stage-1 compilation when typedef count exceeds limit.


ChAoSUnItY (approved):
I've cherry picked this patch and confirmed it was caused by insufficient buffer capacity, which fixes #226.

PR #260, opened 2025-08-25
Use string interning for identifier deduplication

jserv:
This adds string interning to reduce memory usage by deduplicating identical identifier strings throughout the compilation process. It ensures that each unique identifier string is stored only once in memory, with all references pointing to the single interned copy.

The implementation uses a hashmap-based string pool that checks for existing strings before allocating new ones. String interning is now applied comprehensively across all identifier types for maximum memory efficiency.

Benefits:
- Reduces memory usage by 3-5% for typical programs with duplicate identifiers (e.g., common parameter names like 'x', 'y', 'width') 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request implements a string interning mechanism to optimize memory usage by deduplicating identical identifier strings during compilation. A hashmap-based string pool is introduced, enhancing memory efficiency by 3-5% for programs with duplicate identifiers, particularly in 'src/parser.c'.

</div>

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -13,6 +13,9 @@
 
 #include "defs.h"
 
+/* Forward declaration for string interning */
+char *intern_string(char *str);
+
 /* Lexer */
 char token_str[MAX_TOKEN_LEN];
 token_t next_token;
@@ -673,7 +676,8 @@ void add_alias(char *alias, char *value)
             printf("Failed to allocate alias_t\n");
             return;
         }
-        strcpy(al->alias, alias);
+        /* Use interned string for alias name */
+        strcpy(al->alias, intern_string(alias));
         hashmap_put(ALIASES_MAP, alias, al);
     }
     strcpy(al->value, value);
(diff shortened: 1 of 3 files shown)


src/globals.c
```diff
@@ -1119,6 +1162,16 @@ void global_init(void)
...
         arena_alloc(GENERAL_ARENA, MAX_IR_INSTR * sizeof(ph2_ir_t *));
+
+    /* Initialize string pool for identifier deduplication */
+    string_pool = arena_alloc(GENERAL_ARENA, sizeof(string_pool_t));
+    string_pool->strings = hashmap_create(512);
+
+    /* Initialize string literal pool for deduplicating string constants */
+    string_literal_pool =
+        arena_alloc(GENERAL_ARENA, sizeof(string_literal_pool_t));
+    string_literal_pool->literals = hashmap_create(256);
```
ChAoSUnItY, line 1205: It seems hashmap won't be freed after in `global_release`?

jserv: > It seems hashmap won't be freed after in `global_release`?

Yes, the hashmaps for string pools were not being freed in `global_release()`.
Fixed now by adding proper cleanup for both string pool hashmaps.

PR #266, opened 2025-08-26
Implement pointer difference arithmetic

jserv:
This commit adds comprehensive pointer difference support (`ptr1 - ptr2`) that was missing from the existing pointer arithmetic implementation:
- Pointer difference calculation: It computes element count by dividing byte difference by element size (char=1, int=4, etc.)
- Expression support: Enables complex expressions like `*(p + (q - r))`
- Unified element size detection: Consistent logic for both regular and typedef pointers across all arithmetic operations

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -2197,6 +2197,307 @@ bool is_logical(opcode_t op)
     return op == OP_log_and || op == OP_log_or;
 }
 
+/* Helper function to calculate element size for pointer operations */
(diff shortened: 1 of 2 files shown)


jserv:
@cubic-dev-ai review this PR

ChAoSUnItY (approved):
The main logic looks good, only few suggestions made by AI review in `tests/driver.sh` needs to change for better readability.

PR #269, opened 2025-09-01
Introduce Virtual Register (VReg) abstraction layer to improve register allocation efficiency

jserv:
Virtual Register (VReg) abstraction layer is added to improve register allocation efficiency.


src/reg-alloc.c
```diff
@@ -75,6 +106,63 @@ ph2_ir_t *bb_add_ph2_ir(basic_block_t *bb, opcode_t op)
     return n;
 }
 
+int calculate_spill_cost(var_t *var, basic_block_t *bb, int current_idx)
```
DrXiao, line 113: Although the commit message has explained "cost", how about also adding some comments here?

src/reg-alloc.c
```diff
@@ -116,91 +206,96 @@ void load_var(basic_block_t *bb, var_t *var, int idx)
...
-        if (i == operand_0)
-            continue;
-        if (i == operand_1)
-            continue;
-        if (!REGS[i].var)
-            continue;
-        if (REGS[i].var->consumed > REGS[spilled].var->consumed)
-            spilled = i;
+    if (REGS[spilled].var) {
+        vreg_clear_phys(REGS[spilled].var);
```
DrXiao: It can be written in a shorter form.
```c
if (REGS[spilled].var)
    vreg_clear_phys(REGS[spilled].var);
```

PR #300, opened 2025-09-11
Cleanup stale TODO comments

### LINKED ISSUE #275: [BUG] Memory leak: Phi instructions not released in SSA construction
## Description
Memory leak detected in the SSA construction phase where phi instructions are not properly released after use, causing unnecessary memory consumption that grows with program complexity.

## Current Behavior
- Phi instructions allocated during SSA construction remain in memory after they're no longer needed
- Memory usage increases unnecessarily with program size
- Located at `src/ssa.c`

## Expected Behavior
Phi instructions should be properly freed after SSA construction completes to prevent memory leaks.

## Location in Code
- **Primary location**: `src/ssa.c`
- **Function**: `solve_phi_params()`
- **Comment in code**: "FIXME: dangling phi causes memory leak. Should release phi here."

## Impact
- Memory consumption grows unnecessarily with program complexity
- Can affect compilation of large programs
- Reduces efficiency of the self-hosting compiler

## Test Case
```c
// Large program with many control flow paths will show increased memory usage
int complex_function(int n) {
    int result = 0;
    for (int i = 0; i < n; i++) {
        if (i % 2 == 0) {
            result += i;
        } else {
            result -= i;
        }
        // More complex control flow increases phi nodes
        for (int j = 0; j < i; j++) {
            if (j % 3 == 0) {
                result *= 2;
            }
        }
    }
    return result;
}
```

## Proposed Solution
Add proper cleanup code to release phi instructions after they're no longer needed in the SSA construction phase.

## References
- Related to SSA (Static Single Assignment) form optimization
- Affects memory management in `INSN_ARENA`

ChAoSUnItY:
Phi instructions in SSA unit is previously allocated via builtin
malloc, and wasn't freeing after SSA unit completed. Later, we
replaced allocator with arena allocator, which guaranteed its lifetime
is within the compilation, thus the TODO comment is removed as resolved.

Close #275.
    

---

## Summary by cubic
Enable LeakSanitizer in the sanitizer build to catch memory leaks during compiler tests. Adds -fsanitize=leak to CFLAGS/LDFLAGS and removes a stale TODO about freeing SSA phi instruction memory.

### DIFF
--- a/src/ssa.c
+++ b/src/ssa.c
@@ -903,8 +903,6 @@ void bb_unwind_phi(func_t *func, basic_block_t *bb)
         for (phi_operand_t *operand = insn->phi_ops; operand;
              operand = operand->next)
             append_unwound_phi_insn(operand->from, insn->rd, operand->var);
-        /* TODO: Release memory allocated for phi instruction to prevent leaks
-         */
     }
 
     bb->insn_list.head = insn;


Makefile
```diff
@@ -52,8 +52,8 @@ SNAPSHOTS := $(foreach SNAPSHOT_ARCH,$(ARCHS), $(patsubst tests/%.c, tests/snaps
 
 all: config bootstrap
 
-sanitizer: CFLAGS += -fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer -O0
-sanitizer: LDFLAGS += -fsanitize=address -fsanitize=undefined
+sanitizer: CFLAGS += -fsanitize=address -fsanitize=undefined -fsanitize=leak -fno-omit-frame-pointer -O0
```
jserv: Enabling both `-fsanitize=address` and `-fsanitize=leak` is redundant. If your goal is to catch leaks and memory safety issues, just keep `-fsanitize=address`.

jserv:

Issue #299, opened 2025-09-08
Fix compound literals to capture all field values, not just first


jserv:
## Description
Struct compound literals currently only capture the first field value, ignoring subsequent fields. This severely limits their usefulness and violates C99 semantics. Array compound literals also have non-standard behavior that needs correction.

## Current Bug
```c
// Only first field is captured
struct point { int x, y; };
struct point p = (struct point){10, 20};  // y is lost, becomes 0

// Array compound literals broken
int *arr = (int[]){1, 2, 3, 4, 5};  // Non-standard behavior
```

## C99 Standard Reference
**C99 §6.5.2.5 Compound literals**:
> A compound literal provides an unnamed object whose value is given by the initializer list.

The entire initializer list must be processed, not just the first element.

## Root Cause Analysis
In `src/parser.c`, the compound literal parsing only processes one field:
```c
// Current broken implementation
if (is_struct) {
    read_parameter_list_decl();  // Only reads one value!
    // Missing: loop to read all fields
}
```

### 1. Complete Struct Compound Literal Parsing
```c
// src/parser.c - Fix compound literal parsing
void read_compound_literal(type_t *type) {
    lex_expect('{');
    
    if (type->base_type == TYPE_struct) {
        parse_struct_compound_literal(type);
    } else if (type->base_type == TYPE_array) {
        parse_array_compound_literal(type);
    } else {
        error("compound literal requires struct or array type");
    }
    
    lex_expect('}');
}

void parse_struct_compound_literal(type_t *struct_type) {
    // Allocate temporary storage for the struct
    var_t *temp = add_tmp_var(current_block, "compound_lit", struct_type);
    
    int field_index = 0;
    struct_field_t *field = struct_type->fields;
    
    while (!lex_peek_token('}')) {
        if (field_index > 0) {
            if (!lex_accept(',')) {
                break;  // No more initializers
            }
        }
        
        // Check for designated initializer
        if (lex_peek_token('.')) {
            field = parse_designated_field_init(struct_type);
        }
        
        if (!field) {
            error("too many initializers for struct");
            break;
        }
        
        // Parse initializer expression
        read_expr();
        var_t *value = pop_var();
        
        // Generate store to field
        var_t *field_ptr = get_struct_field_ptr(temp, field);
        emit_store(field_ptr, value, field->type);
        
        // Move to next field (if not designated)
        if (!designated) {
            field = field->next;
            field_index++;
        }
    }
    
    // Initialize remaining fields to zero
[...]

PR #309, opened 2025-10-06
Fix array compound literal parsing

lorettayao:
**Summary**
Fix a segfault when evaluating array compound literals like `(int[]){1,2,3,4,5}` in expression context. The literal now yields the temporary array’s address (via decay) instead of collapsing to a single element, so indexing and reads work correctly.

**Motivation**
Previously, code like the snippet below crashed because the array literal was reduced to a scalar and later treated as a pointer.

**Reproduction (manual)**
```c
#include <stdio.h>

int main(void) {
    int *a = (int[]){1,2,3,4,5};   // used to crash later
    printf("Testing basic array compound literal:\n");
    for (int i = 0; i < 5; i++)
        printf("  a[%d] = %d\n", i, a[i]);

    int sum = a[0] + a[4];
    printf("Sum = %d (expect 6)\n", sum);
    return 0;
}
```

Before: segfault
After: prints `a[0..4]` and Sum = 6 as expected

**Approach (high level)**

- Ensure `(type[]){…}` produces a real temporary array object and the expression value decays to its address.
- Only collapse to a scalar when a scalar is actually required by context.
- No changes to struct compound literals or plain scalars.

**Scope**

- This PR addresses the array side of #299 sufficient to eliminate the segfault in basic use.
- Struct compound literals and broader initializer behaviors are intentionally out of scope here.

**Tests**

- No test files are added in this PR. The manual snippet above demonstrates the fix.

**Compatibility**

- Restores standard pointer semantics for array compound literals.
- No behavioral changes for structs or scalars.

**Issue**

- Partially addresses #299 (array compound literals).
    

---
## Summary by cubic
Fix parsing of array compound literals so they allocate a temporary array and decay to its address, preserving pointer semantics and preventing segfaults.

- **Bug Fixes**
  - Eliminates crashes when using (int[]){...} in expressions, assignments, and parameter passing.
  - Correct behavior across binary ops, pointer arithmetic, ternary results, and function-call (incl. variadic) arguments.
  - Zero-length array literals yield 0 for scalar uses to avoid garbage reads.
  - Addresses #299.

- **Refactors**
  - Added parse_array_compound_literal, scalarize_array_literal, and an is_compound_literal flag to centralize handling; scalarize only when a scalar is required.
  - Emits element writes and counts initializers; tracks brace consumption to avoid double reads.
  - Struct and scalar compound literals remain unchanged.
  - Increased MAX_VAR_LEN to 128 to prevent generated temporary name truncation.

<sup>Written for commit 4049889c1289b3cde5a484c4e381d4b3a3ebedf8. Summary will update on new commits.</sup>


src/parser.c
```diff
@@ -1283,6 +1286,42 @@ void parse_array_init(var_t *var,
...
+            count++;
+            if (!lex_accept(T_comma))
+                break;
+            if (lex_peek(T_close_curly, NULL))
+                break;
+        }
+    }
+    lex_expect(T_close_curly);
+    var->array_size = count;
+}
```
jserv, line 1368: Add a new blank line.

src/parser.c
```diff
@@ -1283,6 +1286,71 @@ void parse_array_init(var_t *var,
...
+static bool is_array_literal_temp(var_t *var)
```
jserv: `is_array_literal_temp` is misleading. Refine it.

lorettayao: can you expand on what does misleading mean, that will be helpful, thank you

jserv:
You MUST ensure bootstrapping is fully functional before submitting pull requests.

src/parser.c
jserv: > can you expand on what does misleading mean, that will be helpful, thank you

Check https://github.com/sysprog21/shecc/blob/master/CONTRIBUTING.md carefully.

visitorckw:
IIUC, compound literals are a feature supported only since C99, and the shecc README mentions from the very beginning that this project aims to support ANSI C. Therefore, IMO, this at least does not "fix" anything.

DrXiao:
> IIUC, compound literals are a feature supported only since C99, and the shecc README mentions from the very beginning that this project aims to support ANSI C. Therefore, IMO, this at least does not "fix" anything.

As far as I know, shecc is planned to fully support the C99 standard, so I think the term "fix" is acceptable.

visitorckw:
Fine, but since we haven't claimed to fully support C99 and, IIUC, array compound literals were never supported before, this seems more like supporting a new feature to me, rather than fixing an existing problem.

DrXiao:
In fact, shecc has ability to handle array compound literals, but it **only captures the first element** (#299).

Therefore, this pull request specifically aims to fix it.

visitorckw:
> In fact, shecc has ability to handle array compound literals, but it **only captures the first element** (#299).
> 
> Therefore, this pull request specifically aims to fix it.

Thanks, that resolves my doubt.
However, should we also mention the issue in the first commit?

src/parser.c
```diff
@@ -1283,6 +1286,77 @@ void parse_array_init(var_t *var,
...
+    }
+
+    lex_expect(T_close_curly);
+    var->array_size = count;
+}
+/* Identify compiler-emitted temporaries that hold array compound literals.
+ * Parsing assigns these temporaries synthetic names via gen_name_to (".tN")
+ * and they keep array metadata without pointer indirection.
+ */
+bool is_compound_literal_array_temp(var_t *var)
```
jserv: Rename it to `is_array_literal_placeholder` or a closely related alternative.

lorettayao: changed in latest commit accfedf

visitorckw (changes requested):
Please rebase your branch to keep the git history clean.
Commits that fix problems introduced within the same pull request should be avoided.

src/parser.c
```diff
@@ -1283,6 +1286,77 @@ void parse_array_init(var_t *var,
...
+        elem_size = TY_int->size;
+
+    var_t *scalar = require_typed_var(parent, elem_type);
+    scalar->ptr_level = 0;
+    gen_name_to(scalar->var_name);
+    scalar->init_val = array_var->init_val;
+
+    add_insn(parent, *bb, OP_read, scalar, array_var, NULL, elem_size, NULL);
+    return scalar;
+}
```
jserv, line 1425: Add a line break.

jserv (changes requested):
Use `git rebase -i` to squash commits and refine commit message.

jserv:
> **No issues found** across 7 files

@cubic-dev-ai, Suggest changes to ensure consistent code style and concise, informative comments.

jserv:
I defer to @ChAoSUnItY for confirmation.

src/arm-codegen.c
```diff
@@ -450,7 +450,7 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
             fatal("Unsupported truncation operation with invalid target size");
         }
         return;
-    case OP_sign_ext: {
```
ChAoSUnItY, line 587: Why removing the curly braces while this case branch has variable declaration?

src/parser.c
```diff
@@ -4356,27 +4491,12 @@ basic_block_t *read_body_statement(block_t *parent, basic_block_t *bb)
...
-                        first_elem->type = var->type;
-                        gen_name_to(first_elem->var_name);
-
-                        /* Extract first element from compound literal array */
-                        add_insn(parent, bb, OP_read, first_elem, expr_result,
-                                 NULL, var->type->size, NULL);
-                        expr_result = first_elem;
-                    }
+                    var_t *rhs = expr_result;
+                    if (!var->ptr_level && var->array_size == 0)
```
ChAoSUnItY: How about:
```c
var_t *rhs = opstack_pop();
```

src/parser.c
```diff
@@ -4468,8 +4588,13 @@ basic_block_t *read_body_statement(block_t *parent, basic_block_t *bb)
                     } else {
                         read_expr(parent, &bb);
                         read_ternary_operation(parent, &bb);
+                        var_t *expr_result = opstack_pop();
+                        var_t *rhs = expr_result;
```
ChAoSUnItY: Ditto.

src/arm-codegen.c
lorettayao: I didn’t really mean to drop those braces—they were just left behind while trying other tweaks. but it’s fine either way because C99 lets us declare source_size right after the case label without changing behavior.

ChAoSUnItY: Yes, although C99 allows this. But since in our compiler's standard workflow, stage 0 will be compiled by GCC with `-Wpedantic` compilation option included, and this will cause building process to output warning even in this case it didn't do anything in a harmful way.

lorettayao: I reverted it to the original order in my latest commit

src/parser.c
lorettayao: squashed together in the latest commit!

src/parser.c
lorettayao: squashed together in the latest commit!

jserv:
@lorettayao You can reply to or chat with @cubic-dev-ai so the AI bot can learn from your feedback and provide smarter code reviews over time.

tests/driver.sh
```diff
@@ -724,6 +724,28 @@ int main() {
...
+    for (int i = 0; i < n; i++)
+        s += p[i];
+    return s;
+}
+int main(void) {
+    int s = sum((int[]){1, 2, 3, 0, 0}, 3);
+    return s != 6;
+}
+EOF
+
```
DrXiao, line 757: Add some test cases to validate array compound literals of the `char` and `short` types:
```c
int main(void) {
    char *s = (char[]){'A', 'B', 'C', 'D', 'E'};
    /* ... */
}
```
```c
int main(void) {
    short *s = (short[]){1, 2, 3, 4, 5};
    /* ... */
}
```

DrXiao: Design test cases and verify that the compiler can correctly handle ternary operations involving array compound literals:
```c
int *a = (condition) ? &arr : (int[]){1, 2, 3, 4, 5};
```

lorettayao: test cases added in the latest commit [dbea9ca]

ChAoSUnItY:
The CI error is caused by IR regression test, if you happened to have this test failed, you can execute `make update-snapshots` to update the IR snapshots, then push the updated snapshots so that the test would proceed.

lorettayao:
> The CI error is caused by IR regression test, if you happened to have this test failed, you can execute `make update-snapshots` to update the IR snapshots, then push the updated snapshots so that the test would proceed.

@ChAoSUnItY When i update-snapshots, there are only the fib-riscv.json & hello-riscv.json been updated, is that normal? I wonder if the CI fail is because I did not update fib-arm.json and hello-arm.json.

ChAoSUnItY:
> there are only the fib-riscv.json & hello-riscv.json been updated, is that normal?

Sometimes some changes only affect certain backend, which is normal. If you're not sure if this is correct, you can perform `make check-snapshots` to see the result first on local machine, if it succeeds, then it should be correct on CI, too.

lorettayao:
@jserv I have rebased to master, and I saw my branch is up to date with 'origin/master'. I am confused if the conflict is because of the rebase issue.

visitorckw:
I guess your origin points to your own repository on github instead of the upstream repository under sysprog21?

lorettayao:
> I guess your origin points to your own repository on github instead of the upstream repository under sysprog21?

yes! Thank you for pointing out!

jserv (changes requested):
Use `git rebase -i` to rework commits, squashing intermediate ones.

DrXiao:
@lorettayao, note that the commit message body should be wrapped at 72 characters.

src/parser.c
```diff
@@ -45,7 +45,10 @@ void parse_array_init(var_t *var,
                       block_t *parent,
                       basic_block_t **bb,
                       bool emit_code);
-
+void parse_array_compound_literal(var_t *var,
+                                  block_t *parent,
+                                  basic_block_t **bb,
+                                  bool emit_code);
```
DrXiao: I think this forward declaration is unnecessary here.

src/parser.c
DrXiao: Add a blank line.

src/parser.c
```diff
@@ -1330,6 +1333,85 @@ void parse_array_init(var_t *var,
     }
 }
 
+void parse_array_compound_literal(var_t *var,
+                                  block_t *parent,
+                                  basic_block_t **bb,
+                                  bool emit_code)
```
DrXiao: Is the fourth parameter (`emit_code`) necessary?

src/parser.c
DrXiao: Ditto. Add a blank line.

src/parser.c
```diff
@@ -1330,6 +1333,85 @@ void parse_array_init(var_t *var,
...
+           var->var_name[0] == '.';
+}
+
+bool is_pointer_like_value(var_t *var)
+{
+    return var && (var->ptr_level || var->array_size ||
+                   (var->type && var->type->ptr_level > 0));
+}
+
+var_t *scalarize_array_literal(block_t *parent,
```
DrXiao, line 1390: Add more comments for this function.

src/parser.c
```diff
@@ -3502,17 +3617,41 @@ void read_ternary_operation(block_t *parent, basic_block_t **bb)
...
+
+    if (true_array && !false_ptr_like)
+        true_val = scalarize_array_literal(parent, &then_, true_val,
+                                           false_val ? false_val->type : NULL);
 
-    rs1 = opstack_pop();
-    add_insn(parent, else_, OP_assign, vd, rs1, NULL, 0, NULL);
+    if (false_array && !true_ptr_like)
+        false_val = scalarize_array_literal(parent, &else_, false_val,
+                                            true_val ? true_val->type : NULL);
```
DrXiao: Add comments to explain this handling.

src/parser.c
lorettayao: Yes, I found it unnecessary, will remove

src/parser.c
lorettayao: will remove

src/parser.c
lorettayao: added

src/parser.c
lorettayao: added

src/parser.c
jserv: ?

src/parser.c
```diff
@@ -1330,6 +1332,98 @@ void parse_array_init(var_t *var,
...
+{
+    return var && var->array_size > 0 && !var->ptr_level &&
+           var->var_name[0] == '.';
+}
+
+bool is_pointer_like_value(var_t *var)
+{
+    return var && (var->ptr_level || var->array_size ||
+                   (var->type && var->type->ptr_level > 0));
+}
```
jserv, line 1383: Add a proper line break.

src/parser.c
jserv: ?

src/parser.c
```diff
@@ -1622,7 +1716,14 @@ void read_func_parameters(func_t *func, block_t *parent, basic_block_t **bb)
...
 
         param = opstack_pop();
-
+        if (func) {
+            if (param_num < func->num_params) {
+                var_t *target = &func->param_defs[param_num];
+                if (!target->ptr_level && !target->array_size)
+                    param = scalarize_array_literal(parent, bb, param,
+                                                    target->type);
+            }
```
jserv: Avoid overly deep nesting of `if` statements.

src/parser.c
```diff
@@ -45,7 +45,9 @@ void parse_array_init(var_t *var,
                       block_t *parent,
                       basic_block_t **bb,
                       bool emit_code);
-
+void parse_array_compound_literal(var_t *var,
+                                  block_t *parent,
+                                  basic_block_t **bb);
```
DrXiao: Remove this declaration.

DrXiao (commented):
The second commit (a4aba54) modified both the C source code and the test suites, but its message only describes the latter. The message does not accurately reflect the changes made.

lorettayao:
> I notice that the term "decay" appears in the commit messages, but I am not sure why it is used. Could you explain the reason behind this choice?

i use the word decay to specify fixes around treating array compound literals this way—e.g., assigning (int[]){1,2,3} to an int * or passing it to a function should transparently yield a pointer

jserv (changes requested):
The work touches too many places with similar boilerplate:
```
  if (!var->ptr_level && var->array_size == 0)
      rhs = scalarize_array_literal(parent, &bb, rhs, var->type);
```

This pattern appears in:
- read_body_assignment (line 3815)
- read_body_statement assignments (line 4563)
- read_body_statement nv assignments (line 4661)
- read_ternary_operation twice

Problem: The special case logic is multiplying, not disappearing.

src/parser.c
```diff
@@ -1330,6 +1329,100 @@ void parse_array_init(var_t *var,
...
+}
+/* Identify compiler-emitted temporaries that hold array compound literals.
+ * Parsing assigns these temporaries synthetic names via gen_name_to (".tN")
+ * and they keep array metadata without pointer indirection.
+ */
+bool is_array_literal_placeholder(var_t *var)
+{
+    return var && var->array_size > 0 && !var->ptr_level &&
+           var->var_name[0] == '.';
+}
```
jserv, line 1377: Detecting temporaries by checking if name starts with . is fragile. This is pattern-matching on internal naming conventions rather than explicit type tracking.

Better approach: Add an explicit is_compound_literal flag to `var_t`.

src/parser.c
```diff
@@ -1622,17 +1715,19 @@ void read_func_parameters(func_t *func, block_t *parent, basic_block_t **bb)
         read_ternary_operation(parent, bb);
 
         param = opstack_pop();
-
+        if (func && param_num < func->num_params) {
+            var_t *target = &func->param_defs[param_num];
+            if (!target->ptr_level && !target->array_size)
+                param =
+                    scalarize_array_literal(parent, bb, param, target->type);
+        }
```
jserv, line 1739: This only scalarizes for defined params, but variadic args (func->va_args) might still misbehave. Test case needed:
```
  printf("%p", (int[]){1,2,3});  // Should print pointer, not 1
```

src/parser.c
```diff
@@ -1330,6 +1329,100 @@ void parse_array_init(var_t *var,
...
+            if (lex_peek(T_close_curly, NULL))
+                break;
+        }
+    }
+
+    lex_expect(T_close_curly);
+    var->array_size = count;
+}
+/* Identify compiler-emitted temporaries that hold array compound literals.
+ * Parsing assigns these temporaries synthetic names via gen_name_to (".tN")
```
jserv: Add explicit compound literal tracking instead of `.tN` name inspection.

jserv:
Consider test case below
```c
  // Variadic function with array literal
  int main(void) {
      int *p = (int[]){42, 43, 44};
      return *p == 42 ? 0 : 1;
  }
```

The existing tests cover basic pointer decay but not all edge cases from the cubic analysis.

src/parser.c
lorettayao: Added the test case taking care this in the latest commit

src/parser.c
lorettayao: I added is_compound_literal flag to var_t in defs.h and and switched is_array_literal_placeholder to rely on it instead of the synthetic .t naming.

lorettayao:
I encountered a CI failure with "buffer overflow detected" during the stage 1 bootstrap.

Debugging with AddressSanitizer revealed a stack-buffer-overflow in src/parser.c where a generated temporary name (34 bytes) exceeded the MAX_VAR_LEN limit (32 bytes).

I have increased MAX_VAR_LEN to 128 to accommodate these generated identifiers and prevent similar overflows in strcpy operations. Please let me know if you would prefer a local fix instead of this global change.

jserv:
> Please let me know if you would prefer a local fix instead of this global change.

Instead of asking for decisions, you should present the relevant considerations from your own perspective, since you are both the implementer and the domain expert for the above code.

jserv:
I defer to @ChAoSUnItY for confirmation.

src/parser.c
```diff
@@ -1330,6 +1330,115 @@ void parse_array_init(var_t *var,
...
+     */
+    type_t *result_type = hint_type ? hint_type : literal_type;
+    if (!result_type)
+        result_type = TY_int;
+
+    /* Create a new scalar temporary, giving it a unique name and copying over
+     * the literal data so downstream code can treat it like a normal value.
+     */
+    var_t *scalar = require_typed_var(parent, result_type);
+    scalar->ptr_level = 0;
```
ChAoSUnItY, line 1417: Unnecessary assignment as `require_typed_var` guarantees `ptr_level` to be 0. Remove this assignment.

ChAoSUnItY (commented):
The overall logic looks fine.

src/parser.c
lorettayao: yes, i removed this one

src/parser.c
DrXiao: Add a blank line.

src/parser.c
DrXiao: Is this assignment still necessary?

src/parser.c
DrXiao: Has the required test case been added to `tests/driver.sh`?

src/parser.c
lorettayao: i have removed in th latest commit since require_typed_var already zeroes it

jserv:
Thank @lorettayao for contributing!

Issue #221, opened 2025-07-15
Incorrect behavior of 'error' function


ChAoSUnItY:
`error` function heavily relies on `source_idx`, which results in the function itself not being able to be used after parsing phase completed, as first saw in #220.

https://github.com/sysprog21/shecc/blob/11785eb7a0fe0d7c549b3c755208bbf6a874e28f/src/globals.c#L1123-L1144

## Possible fixes

One possible fix to address this issue is to instead of relying on global counter (`source_idx`), we stores lexical position within an AST's node, and by passing an node instance to properly report an error at correct lexical position.

## Workarounds

It's advised to use `fatal` function after parsing phase is completed before behavior of `error` function is fixed.

ChAoSUnItY:
The initial introduction of #313 resolved this issue, that is, now tokens will persist and retain the original source code information, which allows error message before SSA optimization / register allocation / code generation being more informative and stable.

PR #313, opened 2025-11-15
Major frontend refactor

ChAoSUnItY:
This patch aims to resolve most current preprocessor issues seen in parser unit, including:
1. Inappropriate eager file inclusion expansion, this is caused due to file inclusion is always happening before actual if-else file guard can determine, which sometimes causes weird result.
2. Bad error message reporting system, current system uses source index to suggest user where the error is, which is not really helpful on pratical term.
3. Lack of expansion only compilation.
4. Lack of predefined macros, including `__LINE__`, `__FILE__`.

In this approach, we introduce a whole new phase dedicated to preprocessing (this practice can be seen in chibicc and other similar cc), instead of binding the preprocessing phase in few different places. 

## Current status
This draft is still in development, the below screenshot is comparison between result of `cpp -E` and `out/shecc -E`:

<details>

<summary>comparison</summary>

`test.c`:
```c
#define B
#define EXP(...) __VA_ARGS__
#include "a.c"

int main()
{
#if defined(B)
    __LINE__;
    __FILE__;
#endif
    EXP(1, +, 2);
    printf("Hello, World!\n");
    return 0;
}
```
`a.c`:
```c
#pragma once
int fib(void)
{
    A;
    __FILE__;
    return 0;
}
```

image

</details>

Current major compilation workflow hasn't affected by the patch, which is expected to resolve in later commits.

## Memory overhead
Memory usage increasement is expected since we now use `token_t` struct to track position, and token stream will be generated at once for preprocessor and parser to use with. Once parser finished, we expect to free all tokens.

---
## Summary by cubic
Introduces a dedicated preprocessing phase and rebuilds token handling to fix include guard behavior, add robust macro expansion, and improve error diagnostics. Adds an -E mode to output preprocessed code comparable to cpp -E.

- **New Features**
  - Standalone preprocessor with object/function-like and variadic macros.
  - Conditional directives: #if, #ifdef, #ifndef, #elif, #else, #endif.
  - Local #include support and #pragma once; parses <...> but system paths aren’t resolved yet.
  - Predefined macros: __LINE__ and __FILE__.
  - -E expansion-only mode with preprocessed output emitter.
  - Better token tracking and caching, with precise file:line:column locations.
  - CI job to preprocess and build a stage1 artifact to validate -E output.
  - libc: add ctype predicates (isdigit, isalpha, isalnum, isxdigit, isblank).
  - libc: add fseek/ftell with lseek support for ARM/RISC-V.

- **Bug Fixes**
  - Include expansion now respects guards; files are included after condition evaluation.
  - Error messages upgraded from raw source indices to readable diagnostics with caret highlights.
  - Corrected #elif evaluation during preprocessing.
  - Deferred string/char literal unescaping until after preprocessing; fixes hex and octal escape handling.

<sup>Written for commit 0bb0033be46e3c5c623a811acfbdd4247320db71. Summary will update on new commits.</sup>


src/globals.c
```diff
@@ -1354,19 +1356,334 @@ void global_release(void)
...
 
+    hashmap_free(SRC_FILE_MAP);
+    hashmap_free(MACROS_MAP);
     hashmap_free(FUNC_MAP);
     hashmap_free(INCLUSION_MAP);
     hashmap_free(ALIASES_MAP);
     hashmap_free(CONSTANTS_MAP);
 }
 
+void dbg_token(token_t *token)
```
jserv: Is it needed? Use conditional build.

src/lexer.c
```diff
@@ -239,10 +246,984 @@ char peek_char(int offset)
...
+    return buf->elements[buf->size + offset];
+}
+
+char read_offset(strbuf_t *buf, int offset)
+{
+    buf->size += offset;
+    return buf->elements[buf->size];
+}
+
+char read(strbuf_t *buf)
```
DrXiao: How about renaming it to something else? `read` should be reserved for the system call in C programs.

src/globals.c
```diff
@@ -1209,17 +1210,19 @@ void global_init(void)
     string_literal_pool->literals = hashmap_create(256);
 
     SOURCE = strbuf_create(MAX_SOURCE);
+    SRC_FILE_MAP = hashmap_create(8);
```
DrXiao: How about defining a new macro for the constant value `8`?

src/globals.c
DrXiao: Would it be better to use a global array to store the strings and use `token->kind` as the index to retrieve the corresponding one?
```c
/* Initialize this array in global_init() */
char *token_str[NUM_OF_TOKENS];
```
```c
if (token->kind >= 0 && token->kind <= T_inclusion_path)
    name = token_str[token->kind);
else
    name = "<unknown>";
```

src/lexer.c
ChAoSUnItY: The naming is still under adjustment since I don't want to cause some confliction to the existed functions, which might breaks self-hosting workflow.

ChAoSUnItY:
The initial result is benchmarked below, the memory overhead and minor performance overhead are expected to be the outcome of algorithm and data structure used in this patch:
1. The source files will be loaded once into memory (which used for later error diagnositc and tokenization)
2. The token stream will later be computed (or tokenized) and store in `TOKEN_ARENA`, and this is the major memory overhead expected to be produced in this patch
3. The computed token stream of each source files will also get cached
4. Later in preprocessor, each token will also get copied to be preprocessed, this is to prevent corrupted token integrity if later cached token stream is used. This is also expected to produce another major memory overhead.
5. After the preprocessing is done, comipler will pass the whole computed token stream to parser

The performance overhead is expected to be came from multiuple token stream traversal, as seen 3 times at least in lexer, preprocessor, and parser.

<details>

<summary>performance benchmark (stage 0)</summary>

## Before
`/bin/time -v out/shecc src/main.c`:
```
        Command being timed: "./out/shecc ./src/main.c"
        User time (seconds): 0.16
        System time (seconds): 0.33
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.50
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 305664
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 76195
        Voluntary context switches: 0
        Involuntary context switches: 2
        Swaps: 0
        File system inputs: 0
        File system outputs: 728
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

## After
`/bin/time -v out/shecc src/main.c`:
```
        Command being timed: "./out/shecc ./src/main.c"
        User time (seconds): 0.20
        System time (seconds): 0.44
        Percent of CPU this job got: 98%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.66
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 384000
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 3
        Minor (reclaiming a frame) page faults: 96004
        Voluntary context switches: 44
        Involuntary context switches: 2
        Swaps: 0
        File system inputs: 1992
        File system outputs: 800
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

</details>
<details>

<summary>performance benchmark (stage 1)</summary>

## Before
`/bin/time -v out/shecc-stage1.elf src/main.c`:
```
        Command being timed: "./out/shecc-stage1.elf ./src/main.c"
        User time (seconds): 1.49
        System time (seconds): 1.23
        Percent of CPU this job got: 98%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:02.76
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 188772
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 64
        Minor (reclaiming a frame) page faults: 44474
        Voluntary context switches: 158
        Involuntary context switches: 5
        Swaps: 0
        File system inputs: 10880
        File system outputs: 728
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

## After
`/bin/time -v out/shecc-stage1.elf src/main.c`:
```
        Command being timed: "./out/shecc-stage1.elf ./src/main.c"
        User time (seconds): 1.94
        System time (seconds): 1.50
        Percent of CPU this job got: 99%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:03.47
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 233612
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 63
        Minor (reclaiming a frame) page faults: 55761
        Voluntary context switches: 50
        Involuntary context switches: 15
        Swaps: 0
        File system inputs: 9120
        File system outputs: 800
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```
</details>

ChAoSUnItY:
There are still some improvements left to do, e.g.:
1. Delay string / character literal escaped character computation, so that when passing `-E` the result don't have to convert back again.
2. Function & structure renaming, some leftover unused structures will be also removed as well.

ChAoSUnItY:
I seem to unable to reply to the latest review suggestion so I'll reply here:

> Would it be better to use a global array to store the strings and use `token->kind` as the index to retrieve the corresponding one?

Perhaps? The approach requires reorganization of token kind order, and I don't think it's necessary to just boost performance while debugging (I didn't encoutered significant performance overhead when calling it)?

src/globals.c
ChAoSUnItY: I've guarded this function with conditional macro namely `DEBUG_BUILD` defined in `defs.h`.

ChAoSUnItY:
I'm thinking an approach to validate our compiler's expansion only compilation flag (aka `-E`), the procedure is as follow:
1. `make distclean config out/shecc`
2. `./out/shecc -E src/main.c > out/out.c`
3. `./out/shecc --no-libc -o out/shecc-stage1.elf out/out.c`
4. If the above procedure succeeded, we may continue to bootstrap stage 2 executable with the above procedure as well

@jserv what do you think?

jserv:
> I'm thinking an approach to validate our compiler's expansion only compilation flag (aka `-E`), the procedure is as follow:
> 1. `make distclean config out/shecc`
> 2. `./out/shecc -E src/main.c > out/out.c`

Furthermore, we can adapt the preprocessor implementation from slimcc.

src/globals.c
```diff
@@ -1354,53 +1362,342 @@ void global_release(void)
...
+    for (offset = pos; offset >= 0 && src->elements[offset] != '\n'; offset--)
         ;
 
     start_idx = offset + 1;
 
-    for (offset = 0;
-         offset < MAX_SOURCE && (start_idx + offset) < SOURCE->size &&
-         SOURCE->elements[start_idx + offset] != '\n';
+    for (offset = start_idx;
+         offset < src->capacity && src->elements[offset] != '\n' &&
```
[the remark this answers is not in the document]
ChAoSUnItY: The `src->size` is reused by lexer to traverse again the whole string buffer, which does not actually reflect the occupied size of string buffer, instead, `src->cap` used here actually equals to the actual used size of string buffer.

src/main.c
```diff
@@ -84,11 +92,29 @@ int main(int argc, char *argv[])
...
+    if (libc) {
         libc_generate();
+        libc_token_stream = include_libc();
+    }
+
+    token_stream = lex_token_by_file(in);
+
+    /* concat libc's and input file's token stream */
+    if (libc) {
+        libc_token_stream->tail->next = token_stream->head;
```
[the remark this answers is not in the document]
ChAoSUnItY: The input file's token stream is guaranteed not to be mutated by other functions, which is ok to be concatenated with libc's token stream.

ChAoSUnItY:
#221 is related to this patch, as this patch also fixes error diagnostic context issue. We may also close it upon merged into master branch.

.github/workflows/main.yml
```diff
@@ -56,6 +56,45 @@ jobs:
           make check-sanitizer || exit 1
           make check || exit 1
 
+  preprocessor-host:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        compiler: [gcc, clang]
+        architecture: [arm, riscv]
```
jserv, line 96: Why do you create a new target?

ChAoSUnItY: The idea would be testing preprocessing produces correct output without interfering with the original target, which has additional checks to be performed such as sanitized build, IR regression test, and such.

jserv: If the C preprocessor fails, there is no need to validate the C compiler itself.

ChAoSUnItY: If I understand correctly, stripping out stage 2 building tasks should be enough?

jserv: > If I understand correctly, stripping out stage 2 building tasks should be enough?

It fairly makes sense.

lib/c.c
```diff
@@ -584,6 +590,30 @@ int fputc(int c, FILE *stream)
...
 
+int fseek(FILE *stream, int offset, int whence)
+{
+#if defined(__arm__)
+    __syscall(__syscall_lseek, stream, offset, whence);
+#elif defined(__riscv)
+    /* No need to offset */
+    __syscall(__syscall_lseek, stream, 0, offset, NULL, whence);
+#endif
+    return 0;
```
DrXiao: Handle the `#else` case in the same manner as `ftell()` does.

src/globals.c
```diff
@@ -1354,53 +1362,358 @@ void global_release(void)
...
+}
 
-    /* TODO: Implement line/column tracking for precise error location
-     * reporting. Current implementation only shows source position offset.
-     */
-    printf("[Error]: %s\nOccurs at source location %d.\n%s\n", msg,
-           SOURCE->size, diagnostic);
+/* Reports an error and specifying a position */
+/* FIXME: This function has been deprecated, use `error_at` instead. */
+void error(char *msg)
```
DrXiao: If `error()` is deprecated, would it be better to directly replace all `error()` calls with `error_at()`?

src/globals.c
```diff
@@ -769,20 +668,137 @@ char *intern_string(char *str)
...
+                output[j++] = '"';
+                i++;
+                break;
+            case '?':
+                output[j++] = '\?';
+                i++;
+                break;
+            case 'x': {
+                /* Hexadecimal escape sequence: \xhh */
+                i++;  // Skip 'x'
```
DrXiao: Use C-style comment: `/* Skip 'x' */`.

src/globals.c
```diff
@@ -769,20 +668,137 @@ char *intern_string(char *str)
...
+        } else {
+            /* Regular characters */
+            output[j++] = input[i++];
+        }
     }
-    return 0;
+
+    output[j] = '\0';
+
+    // Check if we ran out of output space
```
DrXiao: Ditto. Use C-style comment.

src/globals.c
```diff
@@ -1208,19 +1224,12 @@ void global_init(void)
...
-    /* Initialize token management globals */
-    current_location.line = 1;
-    current_location.column = 1;
-    current_location.filename = NULL;
-    TOKEN_POOL = NULL;
-    TOKEN_BUFFER = NULL;
-    ALIASES_MAP = hashmap_create(MAX_ALIASES);
     CONSTANTS_MAP = hashmap_create(MAX_CONSTANTS);
 
+    /* Initialize token management globals */
```
DrXiao: Is this comment related to the code that follows?
The `elf_*` variables are involved in the ELF file generation process.

src/globals.c
```diff
@@ -1354,53 +1362,358 @@ void global_release(void)
...
+    hashmap_free(TOKEN_CACHE);
+    hashmap_free(SRC_FILE_MAP);
     hashmap_free(FUNC_MAP);
-    hashmap_free(INCLUSION_MAP);
-    hashmap_free(ALIASES_MAP);
     hashmap_free(CONSTANTS_MAP);
 }
 
+#if DEBUG_BUILD
+void dbg_token(token_t *token)
```
DrXiao: I am not sure whether it would be better to move this function into a new C file, such as `debug.c`.

src/defs.h
```diff
@@ -10,6 +10,18 @@
 
 /* definitions */
 
+#define DEBUG_BUILD false
```
jserv: Remove this.

src/globals.c
ChAoSUnItY: As @jserv suggests removing macro `DEBUG_BUILD`, I might actually remove the function as this function isn't really used anywhere, and was only intended to debug while implementing preprocessor itself.

src/globals.c
ChAoSUnItY: The best way to handle this is to redirects all `error()` usage to either `error_at(...)` or `fatal()`, since there are some certain call sites that unable to fetch the cause of location at this moment.

ChAoSUnItY:
It seems all static test stuite failed on stage 2 compilation after merge conflicts resolution and minor change introduced, which is weird.

DrXiao:
After adding the `-strace` flag, the output shows that `lseek()` reports a "Bad file descriptor" error.

```
$ qemu-arm -strace out/shecc-stage1.elf --no-libc -o test test.c
...
11316 lseek(-9,0,SEEK_CUR) = -1 errno=9 (Bad file descriptor)
11316 lseek(-9,0,SEEK_CUR) = -1 errno=9 (Bad file descriptor)
11316 lseek(-9,0,SEEK_CUR) = -1 errno=9 (Bad file descriptor)
11316 lseek(-9,0,SEEK_CUR) = -1 errno=9 (Bad file descriptor)
11316 lseek(-9,0,SEEK_CUR) = -1 errno=9 (Bad file descriptor)
11316 lseek(-9,0,SEEK_CUR) = -1 errno=9 (Bad file descriptor)
```

lib/c.c
```diff
@@ -543,6 +543,34 @@ int fputc(int c, FILE *stream)
     return c;
 }
 
+int fseek(FILE *stream, int offset, int whence)
+{
+#if defined(__arm__)
+    __syscall(__syscall_lseek, stream, offset, whence);
+#elif defined(__riscv)
+    /* No need to offset */
+    __syscall(__syscall_lseek, stream, 0, offset, NULL, whence);
```
DrXiao: The current `fseek()` implementation does not return a value. It would be better to return an appropriate result.

lib/c.c
```diff
@@ -543,6 +543,34 @@ int fputc(int c, FILE *stream)
...
+    int result;
+    __syscall(__syscall_lseek, stream, 0, 0, &result, SEEK_CUR);
+    return result;
+#else
+#error "Unsupported ftell support for current platform"
+#endif
+}
+
+/* Non-portable: Assume page size is 4KiB */
+#define PAGESIZE 4096
```
DrXiao: `c.h` already contains the same definition, so these lines can be removed.

lib/c.c
```diff
@@ -543,6 +543,33 @@ int fputc(int c, FILE *stream)
...
 
+int fseek(FILE *stream, int offset, int whence)
+{
+#if defined(__arm__)
+    __syscall(__syscall_lseek, stream, offset, whence);
+    return 0;
+#elif defined(__riscv)
+    /* No need to offset */
+    __syscall(__syscall_lseek, stream, 0, offset, NULL, whence);
+    return 0;
```
DrXiao: Does the implementation ensure that `fseek()` always succeeds?

src/defs.h
```diff
@@ -10,6 +10,16 @@
 
 /* definitions */
 
+/* Common macro functions */
+#define is_whitespace(c) (c == ' ' || c == '\t')
+#define is_newline(c) (c == '\r' || c == '\n')
+#define is_alnum(c)                                      \
+    ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || \
+     (c >= '0' && c <= '9') || (c == '_'))
```
DrXiao: `isalnum()`  is equivalent to `isalpha() || isdigit()`, which means the macro should only validate whether the character is a digit or an alphabetic letter.

.github/workflows/main.yml
```diff
@@ -88,6 +88,34 @@ jobs:
           make check-sanitizer DYNLINK=${{ steps.determine-mode.outputs.DYNLINK }} || exit 1
           make check DYNLINK=${{ steps.determine-mode.outputs.DYNLINK }} || exit 1
 
+  preprocessor-host:
```
DrXiao, line 91: Is this target still necessary?

DrXiao (commented):
The following snapshots should be removed.
* `fib-arm.json`
* `fib-riscv.json`
* `hello-arm.json`
* `hello-riscv.json`

src/defs.h
```diff
@@ -10,6 +10,16 @@
...
+#define is_digit(c) ((c >= '0' && c <= '9'))
+#define is_hex(c) \
```
DrXiao: How about renaming them to `isalnum()`, `isdigit()`, and `isxdigit()`, respectively?

Since they are standard C functions/macros, would it be better to move them to `c.h`?

src/globals.c
```diff
@@ -69,6 +52,10 @@ arena_t *BLOCK_ARENA;
 /* BB_ARENA is responsible for basic_block_t / ph2_ir_t allocation */
 arena_t *BB_ARENA;
 
+/* TOKEN_ARENA is responsible for token_t (including literal) /
+ * source_location_t allocation */
```
DrXiao: nitpick:
```
/* TOKEN_ARENA is responsible for token_t (including literal) /
 * source_location_t allocation.
 */
```

.github/workflows/main.yml
ChAoSUnItY: This target is meant to ensure that by compile with shecc with expanded-only compliation flag, then compile with shecc should success, which serves as black box test to ensure that macro expansion is correct.

ChAoSUnItY:
Current `fseek` and `ftell` lack of capability to correctly store state of file descriptor, I'm expected this issue to be resolved outside of this patch, I'll create another issue to track this.

src/defs.h
```diff
@@ -10,6 +10,15 @@
 
 /* definitions */
 
+/* Common macro functions */
+#define is_whitespace(c) (c == ' ' || c == '\t')
+#define is_newline(c) (c == '\r' || c == '\n')
+#define is_alpha(c) ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z'))
+#define is_digit(c) ((c >= '0' && c <= '9'))
+#define is_alnum(c) (is_alpha(c) || is_digit(c))
+#define is_hex(c) \
```
DrXiao: How about renaming them to `isalpha`, `isdigit`, `isalnum`, and `isxdigit`, respectively?

Since these are standard C functions/macros, it would be more appropriate to move their definitions into `lib/c.h`.

src/globals.c
```diff
@@ -776,20 +675,137 @@ char *intern_string(char *str)
...
+int unescape_string(const char *input, char *output, int output_size)
+{
+    if (!input || !output || output_size == 0)
+        return -1;
+
+    int i = 0, j = 0;
+
+    while (input[i] != '\0' && j < output_size - 1) {
+        if (input[i] == '\\') {
+            i++;
```
DrXiao: Rewrite as follows:
```c
if (input[i] != '\\') {
    /* Regular characters */
    output[j++] = input[i++];
    continue;
}

switch (input[i]) {
    /* ... */
}

```

src/lexer.c
```diff
@@ -145,686 +144,756 @@ void lexer_cleanup()
...
+    while (fgets(buffer, MAX_LINE_LEN, f))
+        strbuf_puts(src, buffer);
+
+    fclose(f);
+    src->elements[len] = '\0';
+    return src;
 }
 
-bool is_numeric(char buffer[])
+strbuf_t *cache_or_read_file(char *filename)
```
DrXiao: Would renaming it to `get_file_buf()` be clearer?

src/lexer.c
```diff
@@ -145,686 +144,756 @@ void lexer_cleanup()
...
+    if (ch == '~') {
+        ch = read_char(buf);
+        token = new_token(T_bit_not, loc, 1);
+        loc->column++;
+        return token;
+    }
+
+    if (ch == '"') {
+        int sz = 0;
+        bool special = false;
```
DrXiao, line 506: If I understand correctly, is `special` used to handle multi-lines strings?

e.g.:
```c
printf("Hello \
        Test a multiple lines string\n");
```

If so, how about adding some comments to explain `special`? Please correct me if I am wrong.

src/lexer.c
```diff
@@ -833,110 +902,184 @@ token_t lex_token_impl(bool aliasing)
...
+    hashmap_put(TOKEN_CACHE, filename, tks);
+    return tks;
 }
 
+token_stream_t *include_libc()
+{
+    token_t head;
+    token_t *cur = &head, *tk;
+    token_stream_t *tks;
+    char *filename = "lib/c.c";
```
DrXiao: When using dynamic linking, only `lib/c.h` is used to provide libc function interfaces and macros to the compiled source code. Directly parsing `lib/c.c` appears incorrect.

src/lexer.c
ChAoSUnItY: `special` is still used as handling escaped character, the logic has been changed to delay escaped character transformation after proceeding to parser. The reason to change the behavior is that `-E` will still need original string.

src/globals.c
```diff
@@ -776,20 +675,174 @@ char *intern_string(char *str)
...
+            return value;
+        }
+        if (i == 1 && (buffer[i] | 32) == 'b') { /* binary */
+            value = 0;
+            i = 2;
+            while (buffer[i]) {
+                char c = buffer[i++];
+                value <<= 1;
+                if (c == '1')
+                    value += 1;
```
DrXiao: Rewrite it as `value += (c == '1')`.

src/lexer.c
```diff
@@ -833,110 +903,184 @@ token_t lex_token_impl(bool aliasing)
...
+        return token;
     }
 
-    /* This only happens when parsing a macro. Move to the token after the
-     * macro definition or return to where the macro has been called.
+    error_at("Unexpected token", loc);
+    return NULL;
+}
+
+token_stream_t *lex_token_by_file(char *filename)
```
DrXiao: How about renaming it to `gen_file_token_stream()` or a similar name?

src/lexer.c
```diff
@@ -833,110 +903,184 @@ token_t lex_token_impl(bool aliasing)
...
-    /* Unreachable, but we need an explicit return for non-void method. */
-    return T_eof;
+    tks = malloc(sizeof(token_stream_t));
+    tks->head = head.next;
+    tks->tail = cur;
+    hashmap_put(TOKEN_CACHE, filename, tks);
+    return tks;
 }
 
+token_stream_t *include_libc()
```
DrXiao: Similarly, how about `gen_libc_token_stream()`?

src/main.c
```diff
@@ -90,10 +97,27 @@ int main(int argc, char *argv[])
...
+        token_stream = libc_token_stream;
+    }
+
+    tk = preprocess(token_stream->head);
+
+    if (expand_only) {
+        emit_preprocessed_token(tk);
+
+        return 0;
```
DrXiao: Remove the blank line. Additionally, would it be more appropriate to use `exit(0)` to terminate execution, as done at L160?

lib/c.h
```diff
@@ -52,13 +58,23 @@
 /* va_list support for variadic functions */
 typedef int *va_list;
 
+/* Character predicate functions */
+
```
DrXiao: Remove the blank line.

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+token_t *file_macro_handler(token_t *tk)
+{
+    token_t *new_tk = copy_token(tk);
+    new_tk->kind = T_string;
+    new_tk->literal = tk->location.filename;
+    memcpy(&new_tk->location, &tk->location, sizeof(source_location_t));
+    return new_tk;
+}
+
+token_t *line_macro_handler(token_t *tk)
```
DrXiao, line 105: Add some comments to explain this function.

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+    snprintf(line, MAX_TOKEN_LEN, "%d", tk->location.line);
+
+    token_t *new_tk = copy_token(tk);
+    new_tk->kind = T_numeric;
+    new_tk->literal = intern_string(line);
+    memcpy(&new_tk->location, &tk->location, sizeof(source_location_t));
+    return new_tk;
+}
+
+typedef struct hide_set {
```
DrXiao, line 121: Add some comments to explain `hide_set`.

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+    for (hide_set_t *tmp; hs;) {
+        tmp = hs;
+        hs = hs->next;
+        free(tmp);
+    }
+}
+
+typedef enum { CK_if_then, CK_elif_then, CK_else_then } cond_kind_t;
+
+typedef struct cond_incl {
```
DrXiao, line 171: Add some comments to explain `cond_incl`.

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+{
+    cond_incl_t *cond = arena_alloc(TOKEN_ARENA, sizeof(cond_incl_t));
+    cond->prev = ci;
+    cond->ctx = CK_if_then;
+    cond->tk = tk;
+    cond->included = included;
+    return cond;
+}
+
+typedef struct preprocess_ctx {
```
DrXiao, line 197: Add some comments to explain `preprocess_ctx`.

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+    default:
+        return 0;
+    }
+}
+
+token_t *pp_get_operator(token_t *tk, opcode_t *op)
+{
+    tk = pp_lex_skip_space(tk);
+
+    switch (tk->next->kind) {
```
DrXiao, line 265: Is `tk->next` always valid?

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+            break;
+        case OP_log_or:
+            lhs = lhs || rhs;
+            break;
+        default:
+            error_at("Unexpected infix token while evaluating constant",
+                     &tk->location);
+        }
+
+        tk = pp_get_operator(tk, &op);
```
DrXiao, line 487: Is this line necessary?

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+token_t *pp_skip_inner_cond_incl(token_t *tk)
+{
+    token_kind_t kind;
+
+    while (tk->kind != T_eof) {
+        kind = tk->kind;
+
+        if (kind == T_cppd_if || kind == T_cppd_ifdef ||
+            kind == T_cppd_ifndef) {
+            tk = pp_skip_inner_cond_incl(tk->next->next);
```
DrXiao, line 515: Is it possible for `tk->next` to be `NULL`?

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+            continue;
+        }
+
+        if (kind == T_cppd_endif)
+            return tk->next->next;
```
DrXiao, line 523: Ditto.

src/preprocessor.c
```diff
@@ -0,0 +1,1223 @@
...
+                    hide_set_union(ctx->hide_set, new_hide_set(tk->literal));
+                expansion_ctx.macro_args = hashmap_create(8);
+
+                tk = pp_lex_next_token(tk, true);
+                while (true) {
+                    if (pp_lex_peek_token(tk, T_open_bracket, false)) {
+                        bracket_depth++;
+                    } else if (pp_lex_peek_token(tk, T_close_bracket, false)) {
+                        bracket_depth--;
+                    }
```
DrXiao: Remove the braces `{ }`.

src/preprocessor.c
```diff
@@ -0,0 +1,1253 @@
...
+        tk = pp_read_constant_expr_operand(tk, &lhs);
+    }
+
+    while (true) {
+        tk = pp_get_operator(tk, &op);
+        current_precedence = pp_get_operator_prio(op);
+
+        if (current_precedence == 0 || current_precedence <= precedence) {
+            break;
+        }
```
DrXiao: Remove the brackets `{ }`.

src/preprocessor.c
```diff
@@ -0,0 +1,1253 @@
...
+        int i = 0;
+        for (; i < tk->location.len; i++)
+            dest[i] = ' ';
+        dest[i] = '\0';
+        return dest;
+    }
+    case T_tab:
+        return "\t";
+    case T_start:
+        // FIXME: Unused token kind
```
DrXiao: Use C-style comment: `/* FIXME: Unused token kind */`

src/preprocessor.c
```diff
@@ -0,0 +1,1253 @@
...
+                expansion_ctx.hide_set =
+                    hide_set_union(ctx->hide_set, new_hide_set(tk->literal));
+                cur->next =
+                    pp_preprocess_internal(macro->replacement, &expansion_ctx);
+                cur = expansion_ctx.end_of_token;
+            }
+
+            tk = pp_lex_next_token(tk, false);
+            continue;
+        }
```
DrXiao, line 761: Could you add more comments to explain the handling?

src/elf.c
```diff
@@ -62,7 +62,7 @@ void elf_generate_header(void)
 {
     /* Check for null pointers to prevent crashes */
     if (!elf_code || !elf_data || !elf_symtab || !elf_strtab || !elf_header) {
-        error("ELF buffers not initialized");
+        fatal("ELF buffers not initialized");
```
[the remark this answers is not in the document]
ChAoSUnItY: `error()` is no longer supported, and `fatal()` is now the correct error reporting function since we don't have proper source location to report this kind of error.

jserv:

PR #320, opened 2026-03-14
Fix Valgrind leak paths in lexer and preprocessor

### LINKED ISSUE #319: Valgrind reports definite leaks in minimal compiler runs
## Summary

Running `shecc` under Valgrind reports reproducible definite leaks in the compiler driver, even for a minimal input program.

This appears to come from:
- heap-allocated `token_stream_t` objects in `gen_file_token_stream()` / `gen_libc_token_stream()`
- heap-allocated predefined `macro_t` objects in `preprocess()`
- cached source buffers stored in `SRC_FILE_MAP` whose values are not released before freeing the hashmap

## Reproduction

Minimal input:

```c
int main(void)
{
    return 0;
}
```

Command:

```sh
cat > /tmp/shecc-minimal.c <<'EOF'
int main(void)
{
    return 0;
}
EOF

valgrind \
  --tool=memcheck \
  --leak-check=full \
  --show-leak-kinds=all \
  --track-origins=yes \
  ./out/shecc --no-libc -o /tmp/minimal.elf /tmp/shecc-minimal.c
```

The same issue is also reproducible without `--no-libc`, but `--no-libc` shows that the leak is not caused by `<stdio.h>` or other libc usage.

## Observed result

Valgrind reports:

- `definitely lost: 512 bytes in 6 blocks`
- `indirectly lost: 34 bytes in 1 blocks`

## Expected result

`definitely lost` and `indirectly lost` should both be 0 for this minimal run.

kust1011:
Closes #319.

## Summary

This patch fixes reproducible Valgrind leaks in minimal compiler runs by:

- moving `token_stream_t` allocations in the lexer to `TOKEN_ARENA`
- moving predefined `macro_t` allocations in the preprocessor to `TOKEN_ARENA`
- freeing cached `SRC_FILE_MAP` values before releasing the hashmap

## Verification

Tested with a minimal input:

```c
int main(void)
{
    return 0;
}
```

Valgrind command:

```sh
valgrind \
  --tool=memcheck \
  --leak-check=full \
  --show-leak-kinds=all \
  --track-origins=yes \
  ./out/shecc --no-libc -o /tmp/minimal.elf /tmp/shecc-minimal.c
```

Before:
- `definitely lost: 512 bytes in 6 blocks`
- `indirectly lost: 34 bytes in 1 blocks`

After:
- `in use at exit: 0 bytes in 0 blocks`
- `All heap blocks were freed -- no leaks are possible`

---
## Summary by cubic
Fixes Valgrind-reported memory leaks in the lexer and preprocessor. Moves allocations to the token arena and frees cached source buffers; result: no leaks and no behavior change.

- **Bug Fixes**
  - Allocate token streams in TOKEN_ARENA instead of the heap.
  - Allocate predefined macros in TOKEN_ARENA.
  - Free values in SRC_FILE_MAP before freeing the map.

<sup>Written for commit 0aa2d0991bec9db2144903003e7552e033359f5e. Summary will update on new commits.</sup>

### DIFF
--- a/src/globals.c
+++ b/src/globals.c
@@ -1412,6 +1412,24 @@ int compact_arenas_selective(int phase_mask)
     return total_saved;
 }
 
+static void free_src_file_map_values(void)
+{
+    if (!SRC_FILE_MAP)
+        return;
+
+    for (int i = 0; i < SRC_FILE_MAP->cap; i++) {
+        if (!SRC_FILE_MAP->table[i].occupied)
+            continue;
+
+        strbuf_t *src = SRC_FILE_MAP->table[i].val;
+        if (!src || src == LIBC_SRC)
+            continue;
+
+        strbuf_free(src);
+        SRC_FILE_MAP->table[i].val = NULL;
+    }
+}
+
 void global_release(void)
 {
     /* Cleanup lexer hashmaps */
@@ -1430,6 +1448,7 @@ void global_release(void)
     arena_free(TOKEN_ARENA);
     arena_free(GENERAL_ARENA); /* free TYPES and PH2_IR_FLATTEN */
     hashmap_free(TOKEN_CACHE);
+    free_src_file_map_values();
     hashmap_free(SRC_FILE_MAP);
     hashmap_free(FUNC_MAP);
     hashmap_free(INCLUSION_MAP);
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -966,7 +966,7 @@ token_stream_t *gen_file_token_stream(char *filename)
         error_at("Internal error, expected eof at the end of file",
                  &cur->location);
 
-    tks = malloc(sizeof(token_stream_t));
+    tks = arena_calloc(TOKEN_ARENA, 1, sizeof(token_stream_t));
     tks->head = head.next;
     tks->tail = cur;
     hashmap_put(TOKEN_CACHE, filename, tks);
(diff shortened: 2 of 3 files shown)


src/globals.c
```diff
@@ -1430,6 +1448,7 @@ void global_release(void)
     arena_free(TOKEN_ARENA);
     arena_free(GENERAL_ARENA); /* free TYPES and PH2_IR_FLATTEN */
     hashmap_free(TOKEN_CACHE);
+    free_src_file_map_values();
     hashmap_free(SRC_FILE_MAP);
```
DrXiao, line 1452: How about moving `hashmap_free()` to the new function? That is:
```c
static void free_src_file_map(void)
{
    if (!SRC_FILE_MAP)
        return;

    for (int i = 0; i < SRC_FILE_MAP->cap; i++) {
        if (!SRC_FILE_MAP->table[i].occupied)
            continue;

        strbuf_t *src = SRC_FILE_MAP->table[i].val;
        if (!src || src == LIBC_SRC)
            continue;

        strbuf_free(src);
        SRC_FILE_MAP->table[i].val = NULL;
    }

    hashmap_free(SRC_FILE_MAP);
}
```

Make this new function release `table[i]->val` entries and `SRC_FILE_MAP` simultaneously.

ChAoSUnItY (commented):
Though the current patch's approach is acceptable, however, this would be cumbuersome if same pattern appears in the future, I would prefer trying to patch this in a way that hashmap accepts an externally defined deallocation function, then invokes when hashmap is being freed so that the we don't have to touch low level implementation of hashmap when externally-allocated objects needs to be deallocated.

That being said, this wouldn't be mandatory for this patch to be approved for merging.

jserv:
superseded by #341

PR #321, opened 2026-03-29
Refine generated instructions for staticly linked programs of Arm32

DrXiao:
Previously, for the Arm architecture, the compiler generated the following sequence at the program entry point for staticly linked executables:
- a 'bl' instruction to call a global initialization function, and then execute the main function.
- a 'b' instruction.
- a sequence of instructions to invoke the '_exit' system all.

After further investigation, since the main function inherently invokes '_exit' upon completion, the entry sequence can be simplified. We now use a single 'b' instruction to perform an unconditional jump to the global initialization and the main routines.

Additionally, the implementation of the '__syscall' function has been corrected. Previously, it modified registers r4-r7 without preserving their original values for the caller. This commit ensures that these callee-saved registers are properly saved and restored.

Finally, This refinement improves the efficiency and correctness of the generated instructions for statically linked programs.

---
## Summary by cubic
Simplifies ARM32 static entry by branching directly to global init + main, and fixes `__syscall` to preserve callee-saved registers. Fewer startup instructions and correct ABI behavior.

- **Bug Fixes**
  - `__syscall` now saves/restores r4–r7 with `stmdb/ldm`.
  - Extra args are loaded from the correct stack offsets after the push.

- **Refactors**
  - Replaced the entry sequence with a single `b`; removed explicit `_exit` code.
  - Recomputed ELF offsets to match the shorter static entry path.

<sup>Written for commit fc28ca97d4971580889ea37b8d3fcef07d0607f7. Summary will update on new commits.</sup>

### DIFF
--- a/src/arm-codegen.c
+++ b/src/arm-codegen.c
@@ -165,9 +165,8 @@ void cfg_flatten(void)
             100; /* offset of __libc_start_main + main_wrapper in codegen */
     else {
         func = find_func("__syscall");
-        func->bbs->elf_offset = 60; /* offset of start + exit in codegen */
-        elf_offset =
-            112; /* offset of start + branch + exit + syscall in codegen */
+        func->bbs->elf_offset = 32; /* offset of start + branch in codegen */
+        elf_offset = 92; /* offset of start + branch + syscall in codegen */
     }
 
     GLOBAL_FUNC->bbs->elf_offset = elf_offset;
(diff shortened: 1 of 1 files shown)


src/arm-codegen.c
```diff
@@ -670,30 +669,30 @@ void code_generate(void)
...
-        emit(__mov_r(__AL, __r0, __r0));
-        emit(__mov_i(__AL, __r7, 1));
-        emit(__svc());
+        /* Jump directly to the main preparation and then execute the
+         * main function.
+         *
+         * In static linking mode, when the main function completes its
+         * execution, it will invoke the '_exit' syscall to terminate
+         * the program.
+         */
```
jserv, line 700: Consolidate the comments, illustrating with Arm instructions.

jserv:

PR #322, opened 2026-04-10
Fix wrong op match in peephole optimization

tobiichi3227:
## Summary by cubic
Fixes the peephole optimization to correctly detect and fold a double bitwise NOT. The rule now reduces ~(~x) to x without hitting the arithmetic negate path.

- **Bug Fixes**
  - Match OP_bit_not (not OP_negate) for the "~(~x) => x" rule to prevent incorrect transforms.

<sup>Written for commit 32ba47ab831e46a9b9b3f58717b7cad345ff7264. Summary will update on new commits.</sup>

### DIFF
--- a/src/peephole.c
+++ b/src/peephole.c
@@ -684,7 +684,7 @@ bool bitwise_optimization(ph2_ir_t *ph2_ir)
     /* Pattern 1: Double complement → identity
      * ~(~x) = x
      */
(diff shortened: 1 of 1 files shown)


ChAoSUnItY (approved):

jserv:
Thank @tobiichi3227 for contributing!

PR #323, opened 2026-04-23
Implement RISC-V dynamic linking

DrXiao:
The proposed changes primarily improve the ELF generation and the RISC-V backend, enabling the build system to generate a dynamically linked shecc targeting the RISC-V architecture.

Although the current changes allow both bootstrapping and test suite to complete successfully, this is still a work in progress. The TODO items are listed as follows:
- [x] Enhance code quality, comments and commit messages.
- [x] Confirm the RISC-V ABI compliance.
- [x] Add a new script (`riscv-abi.sh`) to validate the RISC-V ABI.
- [x] Improve the documentation and README.
- [x] Consolidate `arm.mk` and `riscv.mk` into a common build logic (e.g.: configure `RUNNER_LD_PREFIX`).

---
## Summary by cubic
Adds RV32 dynamic linking using RELA with an ABI-compliant PLT/GOT and `__libc_start_main`, and switches `.rel.plt` (Arm) vs `.rela.plt` (RISC‑V) via `ELF_MACHINE_ARM32`/`ELF_MACHINE_RV32`. CI now builds/tests static and dynamic for `arm` and `riscv`, with new RV32 ABI tests, snapshots, and docs.

- **New Features**
  - RISC‑V dynamic linking: `.rela.plt`; PLT0=32B, stubs=16B; `RESERVED_GOT_NUM` (RV32=2, ARM=3); `DYN_LINKER` `/lib/ld-linux-riscv32-ilp32d.so.1`; `ELF_FLAGS` 0x4; correct `DT_RELA*`/`DT_REL*`, `DT_PLTREL`, `DT_PLTRELSZ`, `DT_JMPREL`.
  - Architecture switch: `ELF_MACHINE_*` macros gate `.rel.plt` vs `.rela.plt`.
  - RISC‑V backend/ABI: internal calls direct, extern via PLT; 16‑byte stack alignment; JAL ±1 MB range check; entry via `__libc_start_main`; removed `got_offset`.
  - Build/CI: shared toolchain detection in `mk/common.mk`; Actions download toolchains only for dynamic jobs; matrix runs static+dynamic for `arm` and `riscv`.
  - Tests/Docs: RV32 ABI suite (`tests/riscv-abi.sh`), dynamic snapshots (`hello`, `fib`), expanded dynamic linking docs with RISC‑V PLT/ABI details.

- **Bug Fixes**
  - ABI scripts now validate exit code and output; fixed expected output for multi‑arg printf to avoid false positives.

<sup>Written for commit 6f43590f20230e994dfadb9965958dbd7dfd8140. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/shecc/pull/323?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DrXiao:
As the `apt` package manager only provides a 64-bit RISC-V GNU cross-compilation toolchain, I utilize [riscv-gnu-toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain) , which is a 32-bit variant, and leverage its artifacts for RISC-V dynamic linking development and validation.

The updated GitHub Actions also downloads the 32-bit variant to validate RISC-V dynamic linking.

jserv:
> As the `apt` package manager only provides a 64-bit RISC-V GNU cross-compilation toolchain, I utilize [riscv-gnu-toolchain](https://github.com/riscv-collab/riscv-gnu-toolchain) , which is a 32-bit variant, and leverage its artifacts for RISC-V dynamic linking development and validation.
> The updated GitHub Actions also downloads the 32-bit variant to validate RISC-V dynamic linking.

Evaluate [Run 32-bit applications on 64-bit Linux kernel](https://riscv.org/blog/run-32-bit-applications-on-64-bit-linux-kernel-liu-zhiwei-guo-ren-t-head-division-of-alibaba-cloud/), which is exactly RV32-on-RV64 userspace compatibility, not emulation.

DrXiao:
> Evaluate [Run 32-bit applications on 64-bit Linux kernel](https://riscv.org/blog/run-32-bit-applications-on-64-bit-linux-kernel-liu-zhiwei-guo-ren-t-head-division-of-alibaba-cloud/), which is exactly RV32-on-RV64 userspace compatibility, not emulation.

I'm not sure whether I understand correctly. Do you mean that the proposed changes should be verified on a RISC-V machine?

jserv:
> Do you mean that the proposed changes should be verified on a RISC-V machine?

See https://github.com/sysprog21/kbox/issues/18
The [RISE RISC-V Runners](https://github.com/riseproject-dev/riscv-runner) is a managed GitHub Actions runner service that executes CI/CD workflows on real RISC-V hardware.

DrXiao:
RISE RISC-V Runners' documentation explicitly states that binaries must be compiled for riscv64. According to the [FAQ - What architectures are supported?](https://riseproject-dev.github.io/riscv-runner/docs/faq.html#what-architectures-are-supported).
> RISC-V 64-bit (riscv64) only. All runners execute on physical RISC-V hardware. There is no RISC-V emulation. Binaries must be compiled for riscv64.

I created another branch (`feat/dynlink-for-rv32-test-rv64-runner`) to verify if these RISC-V runners could support riscv32 binaries. However, the [test result](https://github.com/DrXiao/shecc/actions/runs/25591084515/job/75128760590#step:6:1) indicates that the runner failed to execute **statically linked** shecc.

Based on both the documentation and my test, it appears that RISE RISC-V runners lack support for 32-bit executables.

DrXiao:
RISC-V calling convention:

1. First eight arguments are passed to a0-a7, and the extra arguments are pushed onto the stack.
2. stack should be 128-bit (16-byte) aligned. In RISC-V ABIs Specification document, the section 2-1 describes
    > 2-1. Integer Calling Convention
    >>  The stack grows downwards (towards lower addresses) and the stack pointer shall be aligned to a
**128-bit** boundary upon procedure entry.
3. Caller/Callee saved registers:
    | Register ABI Name | Saver  |
    | ----------------- | ------ |
    | zero, gp, tp      | (None) |
    | ra                | Caller |
    | sp                | Callee |
    | a0 - a7           | Caller |
    | s0 - s11          | Callee |
    | t0 - t6           | Caller |

---
### Item 1: is done by the register allocation phase.
The register allocator uses virtual registers (vreg0-vreg7) to allocate reigsters for arguments when encountering a function call. vreg0-vreg7 will be mapped to a0-a7, so first eight arguments are naturally passed to these registers. 

Since the current shecc only supports up to 8 arguments, no extra arguments need to be passed to the stack. Thus, we can skip this handling.

### Item 2: is ensured by the RISC-V code generator.
When handling the stack pointer, the code generator will guarantee that `sp` is always incremented or decremented by a multiple of 16 bytes.

### Item 3:
- ra and sp: are properly handled by the code generator.
- a0 - a7: are implicitly handled in register allocation phase.
- s0 - s11: are not necessary to be preserved or restored.
- t0 - t6: are not necessary to be handled.

Therefore, this item can also be considered complete. Further details and explanations can be found in `riscv-codegen.c`.

DrXiao:
Since there may be additional requirements to address, I am still keeping this pull request as a draft currently, and will continue to make improvements if necessary.

.github/workflows/main.yml
```diff
@@ -27,6 +20,9 @@ jobs:
           sudo apt-get install -q -y qemu-user
           sudo apt-get install -q -y build-essential
           sudo apt-get install -q -y gcc-arm-linux-gnueabihf
+          sudo wget -q https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/2026.04.05/riscv32-glibc-ubuntu-24.04-gcc.tar.xz
```
jserv: Bump to July 15, 2026 release.

DrXiao:
## About `LD_PRELOAD`
The `LD_PRELOAD` mechanism allows specified libraries to be loaded before any other shared libraries (e.g.: libc.so) required by the running process, and the process can call a "wrapper" function.

This raises an interesting question: whether the dynamically linked executables generated by shecc can utilize `LD_PRELOAD` to perform "function interposition'.

Therefore, I prepared the following code and conducted an experiment, and it is pleasing that the generated executables can indeed leverage `LD_PRELOAD` and execute successfully.
### Test library
The 'wrapped' `malloc()` can allocate memory and print the allocated size to standard output.
```c
#include <stddef.h>
#include <string.h>
#include <stdio.h>
#include <dlfcn.h>
#include <unistd.h>
void *malloc(size_t size) {
    char buf[32];
    static void *(*real_malloc)(size_t) = NULL;
    if (real_malloc == NULL) {
        real_malloc = dlsym(RTLD_NEXT, "malloc");
    }
    sprintf(buf, "malloc called, size = %zu\n", size);
    write(2, buf, strlen(buf));
    return real_malloc(size);
}
```

Use `arm-linux-gnueabihf-gcc` to build `libmcount.so` for later use.
```shell
$ arm-linux-gnueabihf-gcc -D_GNU_SOURCE -shared -ldl -fPIC -o libmcount.so malloc_count.c
```

### Test program
```c
/* test.c */
int main()
{
    int *a[3];

    a[0] = malloc(sizeof(int));
    a[1] = malloc(sizeof(int) * 7);
    a[2] = malloc(sizeof(int) * 1024);

    free(a[0]);
    free(a[1]);
    free(a[2]);

    return 0;
}
```

Use shecc to compile a dynamically linked executable, and then run it in QEMU with `LD_PRELOAD=./libmcount.so`.
```shell
$ make DYNLINK=1
$ qemu-arm -L /usr/arm-linux-gnueabihf/ out/shecc-stage2.elf --dynlink -o test test.c
$ qemu-arm -L /usr/arm-linux-gnueabihf/ -E LD_PRELOAD=./libmcount.so ./test
malloc called, size = 4
malloc called, size = 28
malloc called, size = 4096
```
We can observe that `test.c` calls the wrapped `malloc()` three times. Each call successfully allocates memory and correctly prints the allocated size.

---
Function interposition is also supported on the RISC-V architecture:
```shell
$ riscv32-unknown-linux-gnu-gcc -D_GNU_SOURCE -shared -ldl -fPIC -o libmcount.so malloc_count.c
$ make DYNLINK=1 ARCH=riscv
$ qemu-riscv32 -L $(riscv32-unknown-linux-gnu-gcc --print-sysroot) out/shecc-stage2.elf --dynlink -o test test.c
$ qemu-riscv32 -L $(riscv32-unknown-linux-gnu-gcc --print-sysroot) -E LD_PRELOAD=./libmcount.so ./test
malloc called, size = 4
malloc called, size = 28
malloc called, size = 4096
```

---
I will conduct this experiment on real hardware (e.g.: BeagleBone Black) later.

src/riscv-codegen.c
```diff
@@ -482,47 +601,273 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
...
+
+        /* The main wrapper is located here under the dynamic linking mode
+         *
+         * Use t1 and t2 registers to temporarily store 'argc' and 'argv',
+         * while preserving ra on the stack.
+         *
+         * After the main function completes its execution, it must use
+         * the content of ra to transfer control back to __libc_start_main().
+         */
+        emit(__addi(__t1, __a0, 0));
```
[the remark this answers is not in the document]
DrXiao: Global initialization does not utilize `t1` and `t2`. Therefore, we can ignore this situation.

mk/common.mk
```diff
@@ -22,6 +22,47 @@ NO_COLOR = \e[0m
...
+        AVAILABLE_TOOLCHAINS := $(foreach tc, $(TOOLCHAIN_CANDIDATES), $(if $(shell which $(tc)gcc), $(tc)))
+        CROSS_COMPILE := $(firstword $(AVAILABLE_TOOLCHAINS))
+
+        ifndef CROSS_COMPILE
+            $(error "Unable to find a proper GNU toolchain.")
+        endif
+
+        ARCH_CC = $(CROSS_COMPILE)gcc
+
+        LD_LINUX_PATH := $(shell cd $(shell $(ARCH_CC) --print-sysroot) 2>/dev/null && pwd)
```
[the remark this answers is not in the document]
DrXiao: We can assume that the developer's cross compiler is always configured with a sysroot.

src/riscv-codegen.c
jserv: The dynamic-startup wrapper stashes `argc`/`argv` in `t1`/`t2`, then calls `GLOBAL_FUNC` (global initializers) before restoring them to `a0`/`a1` for `main` (lines 707-708). `t1`/`t2` are caller-saved temporaries the backend uses freely (div/mod, address arithmetic), so any nontrivial global initializer clobbers `argc`/`argv` and `main` receives garbage. Save them in the wrapper's stack frame, or use callee-saved `s` registers with proper save/restore.

src/riscv-codegen.c
```diff
@@ -482,47 +601,273 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
...
+        emit(__addi(__a2, __sp, 4));
+        emit(__addi(__a3, __zero, 0));
+        emit(__addi(__a4, __zero, 0));
+        emit(__addi(__a5, __zero, 0));
+        emit(__addi(__a6, __sp, 0));
+
+        /* Call __libc_start_main() via PLT[1] */
+        ofs = (dynamic_sections.elf_plt_start + PLT_FIXUP_SIZE) -
+              (elf_code_start + elf_code->size);
+        emit(__jal(__ra, ofs));
```
jserv, line 664: This `jal` to `PLT[1]` (and the extern calls at OP_call / OP_address_of_func) spans `.text` + `.rodata` + `.rel(a).plt` to reach `.plt`, so the offset is roughly `code_size + rodata_size`. `rv_encode_J` in src/riscv.c has no range check, unlike `rv_encode_I/S/B` which all `fatal()` on overflow. Past the J-type reach of ±1 MB the high bits are silently dropped and the jump lands at the wrong address, a silent miscompile the small hello/fib tests never approach. Add the same guard to `rv_encode_J`: `if (imm > 1048575 || imm < -1048576) fatal("Offset too large");` (src/riscv.c is not part of this PR, so this comment is anchored on the call site.)

mk/riscv.mk
```diff
@@ -7,13 +7,13 @@ ARCH_DEFS = \
     \#define ARCH_PREDEFINED \"__riscv\" /* Older versions of the GCC toolchain defined __riscv__ */\n$\
     \#define ELF_MACHINE 0xf3\n$\
     \#define ELF_FLAGS 0\n$\
-    \#define DYN_LINKER \"/lib/ld-linux.so.3\"\n$\
+    \#define DYN_LINKER \"/lib/ld-linux-riscv32-ilp32d.so.1\"\n$\
```
jserv, line 13: The interpreter is `ld-linux-riscv32-ilp32d.so.1` (double-float ABI) while `ELF_FLAGS` remains `0`, which declares soft-float. The object's `e_flags` should carry `EF_RISCV_FLOAT_ABI_DOUBLE` (0x4) to match the ilp32d loader, or the target should select an ILP32 soft-float linker/toolchain. As written the ABI tag contradicts the chosen interpreter and may be rejected or misinterpreted by strict tools.

src/riscv-codegen.c
```diff
@@ -482,47 +601,273 @@ void emit_ph2_ir(ph2_ir_t *ph2_ir)
...
+         *
+         * Use t1 and t2 registers to temporarily store 'argc' and 'argv',
+         * while preserving ra on the stack.
+         *
+         * After the main function completes its execution, it must use
+         * the content of ra to transfer control back to __libc_start_main().
+         */
+        emit(__addi(__t1, __a0, 0));
+        emit(__addi(__t2, __a1, 0));
+        emit(__sw(__ra, __sp, -4));
```
jserv: `sw ra, sp, -4` writes below `sp` before `sp` is decremented (same idiom as `OP_define` at line 298). RISC-V defines no stack red zone, so this stores outside the current frame; a signal or interrupt in that window corrupts the saved `ra`. Decrement `sp` first, then store `ra` at a positive in-frame offset.

src/elf.c
```diff
@@ -664,118 +700,182 @@ void elf_generate_sections(void)
...
-         * PTR_SIZE * 3, respectively. Therefore, no offset assignment is
-         * required for this function.
+        /* Because PLT[1] and GOT[RESERVED_GOT_NUM] are reserved for
+         * __libc_start_main, its plt_offset and got_offset must be
+         * PLT_FIXUP_SIZE and PTR_SIZE * RESERVED_GOT_NUM, respectively.
+         * Therefore, no offset assignment is required for this function.
          */
 
         func_plt_ofs = PLT_FIXUP_SIZE + PLT_ENT_SIZE;
         func_got_ofs = PTR_SIZE << 2;
```
jserv: `func_got_ofs = PTR_SIZE << 2` hardcodes the ARM GOT layout (first extern at GOT[4]). For RV32 (`RESERVED_GOT_NUM=2`) the first extern lives at GOT[3], so this value is wrong for RISC-V. It is latent today only because `func->got_offset` (assigned below) is write-only, the RV32 PLT computes its slot positionally. Use `PTR_SIZE * (RESERVED_GOT_NUM + 1)`, or drop the dead `got_offset` field.

src/elf.c
```diff
@@ -664,118 +700,182 @@ void elf_generate_sections(void)
...
          */
-        elf_write_int(dynamic_sections.elf_got,
-                      dynamic_sections.elf_got_start +
-                          dynamic_sections.got_size +
-                          dynamic_sections.elf_dynstr->size +
-                          dynamic_sections.elf_dynsym->size);
-        elf_write_int(dynamic_sections.elf_got, 0);
-        elf_write_int(dynamic_sections.elf_got, 0);
-        for (int i = PTR_SIZE * 3; i < dynamic_sections.got_size; i += PTR_SIZE)
+        switch (ELF_MACHINE) {
```
jserv, line 759: GOT init relies on implicit `ELF_MACHINE_ARM32 -> ELF_MACHINE_RV32` case fallthrough. It is correct for both arches (ARM writes GOT[0]=&.dynamic then two zeros; RV32 writes two zeros), but it emits `-Wimplicit-fallthrough`, and a stray `break` or a case reorder would silently corrupt ARM GOT init with no failure on the RV32 CI leg. Add a `/* fall through */` annotation, or write the shared zero entries explicitly per case.

tests/riscv-abi.sh
```diff
@@ -0,0 +1,608 @@
...
+    local run_cmd="${TARGET_EXEC:-}"
+    run_cmd="$run_cmd /tmp/shecc_abi_test_$$.elf"
+
+    local run_output
+    local exit_code
+    run_output=$(eval "$run_cmd" 2>&1)
+    exit_code=$?
+
+    # Check result
+    if [[ $exit_code -eq 0 ]]; then
```
jserv: A test is counted as passed on `exit_code == 0` alone; an `expected_output` mismatch is only a warning, and only under VERBOSE. A broken external-call path that still exits 0 passes silently, which is exactly the failure mode this suite is meant to catch. Treat a non-matching `expected_output` as a failure.

.github/workflows/main.yml
```diff
@@ -26,7 +19,13 @@ jobs:
           sudo apt-get install -q -y graphviz jq
           sudo apt-get install -q -y qemu-user
           sudo apt-get install -q -y build-essential
-          sudo apt-get install -q -y gcc-arm-linux-gnueabihf
+          if [ "${{ matrix.architecture }}" = "arm" ]; then
+            sudo apt-get install -q -y gcc-arm-linux-gnueabihf
+          else
+            sudo wget -q https://github.com/riscv-collab/riscv-gnu-toolchain/releases/download/2026.07.15/riscv32-glibc-ubuntu-24.04-gcc.tar.xz
```
jserv: The riscv32 toolchain tarball is fetched with `sudo wget` and extracted with no checksum or signature verification, so a compromised or swapped release would run unchecked in CI. Pin and verify a SHA256 before extraction rather than trusting the download; also `sudo wget` into the workspace is unnecessary.

jserv (commented):
Review notes on the RV32 dynamic linking implementation. Findings are inlined at their locations.

docs/dynamic-linking.md
```diff
@@ -69,30 +74,69 @@ Low Address
...
+- Since the current implementation of shecc supports up to 8 arguments, no argument needs to be passed onto the stack.
+
+Callee's behavior
+
+- Allocate necessary space on the stack and align the stack pointer to 128-bit (16-byte).
+- Preserve callee-saved registers:
+  - Although `sp` is not explicitly saved onto the stack after allocating space for local variables, the code generator guarantees that `sp` is correctly restored for the caller prior to returning. Therefore, `sp` is not necessary to be additionally handled.
+  - Exception:
+    - `s1` - `s11` are not used by the code generator, so they are unnecessary to be processed.
+    - `s0` is unused during static linking and is only accessed at the program's entry point under dynamic linking, there is no need to save this register.
```
DrXiao: Currently, `s0` and `s1` may be used under dynamic linking. I will update the above descriptions.

src/riscv-codegen.c
DrXiao: I have created a new commit to add the necessary range check to `rv_encode_J()`.

mk/riscv.mk
DrXiao: The current `ELF_FLAGS` is set to `0x04`, and the generated ELF executables will use double-float ABI.

src/elf.c
DrXiao: Two `elf_write_int()` calls and a `break` statement have been explicitly added to resolve the fallthrough issue.

DrXiao:
> ## About `LD_PRELOAD`
> The `LD_PRELOAD` mechanism allows specified libraries to be loaded before any other shared libraries (e.g.: libc.so) required by the running process, and the process can call a "wrapper" function.
> ... 
> I will conduct this experiment on real hardware (e.g.: BeagleBone Black) later.

I have used my BeagleBone Black to conduct the same test, and the `LD_PRELOAD` mechanism still works on real hardware.

Demonstration:

https://github.com/user-attachments/assets/85cf0bea-e2b5-48d7-a6a7-0eb1de2ca89b

jserv:

Issue #314, opened 2025-11-22
Improve large structure instances handling


DrXiao:
## Description
Currently, shecc cannot correctly handle large structure instances, causing operations such as large structure assignment and passing a large structure instance as a function argument to behave incorrect.

## Current Bug
```c
typedef struct {
    int a, b, c, d, e;
} large_struct_t;

void test(large_struct_t s)
{
    printf("test(s) = %d %d %d %d %d\n", s.a, s.b, s.c, s.d, s.e);
}

int main(void)
{
    large_struct_t s;
    s.a = 10;
    s.b = 20;
    s.c = 30;
    s.d = 40;
    s.e = 50;
    printf("s = %d %d %d %d %d\n", s.a, s.b, s.c, s.d, s.e);

    /* Bug 1: assignment is incorrect. */
    large_struct_t s2 = s;
    printf("s2 = %d %d %d %d %d\n", s2.a, s2.b, s2.c, s2.d, s2.e);

    /* Bug 2: pass a large instance as an argument is also incorrect. */
    test(s);
    return 0;
}
```
```
$ qemu-arm out/shecc-stage2.elf -o test test.c
$ qemu-arm test
s = 10 20 30 40 50
s2 = 10 0 0 0 0
test(s) = 10 1082129836 1082129828 1082129596 4
```
## Root Cause
One of potential root causes is that the current register allocator lacks proper considerations about "physical storage". Consider the following structure definition in shecc's source code:
```c
typedef struct {
    var_t *var;
    int polluted;
} regfile_t;
```
shecc uses instances of `regfile_t` to perform register allocation. Although a register may conceptually hold a variable (`var_t`), each register on Arm32 and RISC-V can store only a 4 byte object and thus cannot hold large objects.

However, during register allocation, shecc directly assigns each variable to a register even if the variable is a large structure whose size exceeds 4 bytes. Since no additional handling is performed, the code generator ultimately treats every register as if it always stores small objects when generating instructions, which leads to the aforementioned bugs.

jserv:
The reported large-record assignment and by-value argument failures are fixed and covered by direct and indirect record-call regressions. Validation passed on x64 and Arm in both stage 0 and stage 2 (729 driver tests each), with ABI suites passing.

PR #328, opened 2026-08-25
Fix function pointer assignment

ChAoSUnItY:
This patch fixes function pointer assignment on local variable, direct struct member, and pointer-to-struct member, and also fixes in-register arguments not cleared after invocation of function pointer.

---
## Summary by cubic
Fixes function-pointer assignments so locals and struct members use the correct address and stored pointer value instead of treating variables as function symbols. Indirect calls now clear register mappings afterward, preventing stale arguments from corrupting later calls.

**Bug Fixes**
- Loads function-pointer variables before indirect calls while keeping direct function calls unchanged.
- Handles assignments to local variables, direct struct members, and pointer-to-struct members.
- Copies stored addresses correctly between function-pointer variables.
- Lets a local function pointer shadow a global function instead of resolving to the function symbol.
- Materializes local aggregates containing function-pointer fields in backing storage, including nested aggregates.
- Zero-initializes the type arena so new flags start cleared.
- Adds regression tests for assignments, aggregate addressing, and stale argument-register mappings.

<sup>Written for commit 440d3ae46ab084c902e319457c3345fad5bc124a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/shecc/pull/328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -99,6 +99,7 @@ var_t *require_var(block_t *blk)
     var->base = var;
     var->type = TY_int;
     var->space_is_allocated = false;
+    var->has_backing_storage = false;
     var->ofs_based_on_stack_top = false;
     return var;
 }
@@ -110,6 +111,17 @@ var_t *require_typed_var(block_t *blk, type_t *type)
 
     var_t *var = require_var(blk);
     var->type = type;
+    return var;
+}
+
+/* Function-address operands carry a function name, but are not declarations
+ * in the current scope.  Keeping them out of the local lookup list lets
+ * find_var() distinguish a resolved function-pointer variable from a
+ * generated function symbol. */
+var_t *require_func_symbol_var(block_t *blk)
+{
+    var_t *var = require_var(blk);
+    blk->locals.size--;
     return var;
 }
 
@@ -2062,7 +2074,7 @@ void read_expr_operand(block_t *parent, basic_block_t **bb)
                 add_insn(parent, *bb, OP_func_ret, vd, NULL, NULL, 0, NULL);
             } else {
                 /* indirective function pointer assignment */
-                vd = require_var(parent);
+                vd = require_func_symbol_var(parent);
                 vd->is_func = true;
                 strcpy(vd->var_name, intern_string(token));
                 opstack_push(vd);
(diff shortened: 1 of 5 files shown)


src/parser.c
```diff
@@ -3321,10 +3321,15 @@ bool read_body_assignment(char *token,
         } else if (lex_peek(T_open_bracket, NULL)) {
             /* dereference lvalue into function address */
             rs1 = opstack_pop();
-            vd = require_var(parent);
-            gen_name_to(vd->var_name);
-            opstack_push(vd);
-            add_insn(parent, *bb, OP_read, vd, rs1, NULL, PTR_SIZE, NULL);
+
+            if (lvalue.is_reference) {
```
DrXiao, line 3338: How about adding some comments?

src/parser.c
```diff
@@ -3455,6 +3460,16 @@ bool read_body_assignment(char *token,
             if (lvalue.is_func) {
                 rs2 = opstack_pop();
                 rs1 = opstack_pop();
+
+                if (!lvalue.is_reference) {
```
DrXiao, line 3494: Ditto.

src/parser.c
```diff
@@ -3455,6 +3460,19 @@ bool read_body_assignment(char *token,
             if (lvalue.is_func) {
                 rs2 = opstack_pop();
                 rs1 = opstack_pop();
+
+                /* Aquire destination address of lvalue if lvalue is a
```
DrXiao: Typo: aquire -> acquire

tests/driver.sh
```diff
@@ -1643,6 +1643,138 @@ int main() {
...
+    holder_t *p = &h;
+    return p->fn(p->value);
+}
+
+int main() {
+    holder_t h;
+    h.fn = suc;
+    h.value = 4;
+    return call(h);
+}
```
DrXiao: In this case, `call()` takes an 8-byte structure as a parameter and performs an indirect call through it.

As described in issue #314, shecc currently does not support handling cases like `call()`. Therefore, I believe that `call()` cannot correctly receive a `holder_t` instance, because only the first four bytes of the structure are passed in the first register.

Does the proposed changes allow shecc to handle large structure parameters, either fully or partially?

ChAoSUnItY: Yes, the current proposed changes is trying to resolve this issue in partial, though the patch is keep getting complicated, so I may address this fully in other branch instead. I'll strip large strcuture parameters support from this patch, as the goal is still to fix function pointer assignment.

src/reg-alloc.c
```diff
@@ -747,6 +792,43 @@ void reg_alloc(void)
...
+                     *
+                     * FIXME: This does not support aggregate parameter for now.
+                     */
+                    bool is_pointer =
+                        insn->rs1->ptr_level ||
+                        (insn->rs1->type && insn->rs1->type->ptr_level);
+                    if (!insn->rs1->is_global && !is_pointer &&
+                        aggregate_has_function_pointer(insn->rs1->type)) {
+                        if (!insn->rs1->has_backing_storage) {
+                            insn->rs1->offset = func->stack_size;
```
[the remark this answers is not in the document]
ChAoSUnItY: This should be ignored for now, as this issue is supposed to be resolved in different patch.

PR #335, opened 2026-09-09
Support immediate binding option

DrXiao:
Since the immediate binding has already been implemented for the x86-64 and AArch64 architectures, the proposed changes extend its use by introducing a new "-z" command-line option, allowing the Arm32 and RV32 targets to also generate dynamically linked executables with immediate binding.

For the two 64-bit targets, immediate binding is always used under the dynamic linking mode as is.

The update usage:
```shell
# Support the "-z" option
$ qemu-arm -L /usr/arm-linux-gnueabihf/ out/shecc-stage2.elf
Usage: shecc [-o output] [+m] [--dot] [--dump-ir] [--no-libc] [--dynlink] [-z <lazy | now>] [-E] <input.c>
[Error]: Missing source file

# The build system can use the new 'BINDING' variable to determine whether to use immediate binding.
$ make ARCH=arm DYNLINK=1 BINDING=now
```

TODO: 
- [ ] Update the GitHub Actions workflow to validate builds when using immediate binding.

---
## Summary by cubic
Adds a `-z` option to control dynamic binding: `lazy` (default) or `now` (immediate). It replaces the hardcoded immediate binding on x86-64 and AArch64 and lets Arm32 and RV32 dynamically linked executables use immediate binding.

- The Makefile gains a `BINDING` variable, so builds can pass `-z now` via `make DYNLINK=1 BINDING=now`; the ABI and driver tests now take the binding to cover both modes.
- x86-64 and AArch64 still force `now` in dynamic mode, so `-z lazy` is ignored there.

<sup>Written for commit 3c3998dbcb8a46c38987958314ae9ec19ccefd27. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/shecc/pull/335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jserv (changes requested):
Rebase latest `master` branch and resolve conflicts.

.github/workflows/main.yml
```diff
@@ -21,7 +21,7 @@ jobs:
   # Bootstrap each target and run the consolidated behavioral suite against
   # the stage 0 and stage 2 compilers.
   host-x86:
-    name: ${{ matrix.architecture }}/${{ matrix.link_mode }} (${{ matrix.compiler }})
+    name: ${{ matrix.architecture }}/${{ matrix.link_mode }}/${{ matrix.link_mode == 'dynamic' && (matrix.binding_mode == 'immediate' && 'immediate' || 'default') || 'none' }} (${{ matrix.compiler }})
```
DrXiao, line 24: Naming:
- static linking mode: `<arch>/static/none`.
- dynamic linking mode: `<arch>/dynamic/<default or immediate>`

The default binding mode for each architecture is as follows:
- Arm32: lazy binding
- RISC-V: lazy binding
- x64: immediate binding
- Arm64: immediate binding

DrXiao:
- Static linking mode does not involve lazy or immediate binding.
- Only the Arm32 and RISC-V architectures can test both lazy binding and immediate binding modes.
- x64 and Arm64 always use immediate binding.

These conditions make the workflow modifications somewhat complex.

tests/riscv-abi.sh
```diff
@@ -144,7 +146,7 @@ run_abi_test()
     # Compile
     local compile_cmd="$SHECC"
     if [[ "$DYNLINK" == "1" ]]; then
-        compile_cmd="$compile_cmd --dynlink"
+        compile_cmd="$compile_cmd --dynlink -z $BINDING"
```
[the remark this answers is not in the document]
DrXiao: This script is primarily driven by the `make check-abi-stage*` command, so we can assume that `$BINDING` is always valid, since `make check-abi-stage*` always passes a correct value.

tests/arm-abi.sh
```diff
@@ -144,7 +146,7 @@ run_abi_test()
     # Compile
     local compile_cmd="$SHECC"
     if [[ "$DYNLINK" == "1" ]]; then
-        compile_cmd="$compile_cmd --dynlink"
+        compile_cmd="$compile_cmd --dynlink -z $BINDING"
```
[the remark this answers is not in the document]
DrXiao: This script is primarily driven by the `make check-abi-stage*` command, so we can assume that `$BINDING` is always valid, since `make check-abi-stage*` always passes a correct value.

.github/workflows/main.yml
```diff
@@ -32,10 +32,28 @@ jobs:
         compiler: [gcc, clang]
         architecture: [arm, arm64, riscv, x64]
         link_mode: [static, dynamic]
+        include:
```
jserv, line 35: An `include` entry whose keys all match an existing combination is merged into that combination instead of creating a new job, so these four entries convert the existing `arm`/`dynamic` and `riscv`/`dynamic` jobs into immediate-binding runs rather than adding jobs. The default lazy path then has no dynamic coverage on Arm32 or RV32 at all, and the `include` blocks in `sanitizer`, `host-arm` and `host-arm-sanitizer` behave the same way. Adding `binding_mode: [lazy]` to the matrix makes the `immediate` entries conflict, so they expand into separate jobs and the lazy jobs survive.

.github/workflows/main.yml
```diff
@@ -135,17 +161,22 @@ jobs:
...
     timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
         architecture: [arm, arm64]
         link_mode: [static, dynamic]
+        include:
+          - architecture: arm
+            link_mode: dynamic
+            binding_mode: immedaite
```
jserv, line 175: `immedaite` never matches the `matrix.binding_mode == 'immediate'` test that computes `BINDING`, so this entry merges into the existing `arm`/`dynamic` job and leaves it on lazy binding. The native Arm64 runner is the only place a real loader resolves these relocations, so immediate binding ends up untested where it matters most.

```suggestion
            binding_mode: immediate
```

Makefile
```diff
@@ -236,11 +237,11 @@ uninstall-hooks:
 
 check-stage0: $(OUT)/$(STAGE0) tests/driver.sh
 	$(VECHO) "  TEST STAGE 0\n"
-	tests/driver.sh 0 $(DYNLINK)
+	tests/driver.sh 0 $(DYNLINK) $(BINDING)
 
 check-stage2: $(OUT)/$(STAGE2) tests/driver.sh
 	$(VECHO) "  TEST STAGE 2\n"
-	tests/driver.sh 2 $(DYNLINK)
+	tests/driver.sh 2 $(DYNLINK) $(BINDING)
```
jserv, line 244: `check-stage0` and `check-stage2` forward `$(BINDING)`, but `check-sanitizer` below still calls `tests/driver.sh 0 $(DYNLINK)`, so the sanitizer jobs that set `BINDING=now` run the suite with lazy binding. Pass `$(BINDING)` there as well.

src/elf.c
```diff
@@ -1112,14 +1112,17 @@ void elf_generate_dynamic_sections(void)
...
-#if DYN_BIND_NOW == 1
+
     /* Resolve every PLT entry at load time. This target's PLT[0] does not
      * arrange the GOT[1]/GOT[2] hand-off the lazy resolver needs, so the loader
      * writes the final addresses straight into the GOT instead.
      */
-    elf_write_dyn(dynamic_sections.elf_dynamic, 0x18, 0x0); /* DT_BIND_NOW */
-    elf_write_dyn(dynamic_sections.elf_dynamic, 0x1e, 0x8); /* DF_BIND_NOW */
-#endif
+    if (imm_binding) {
```
jserv, line 1120: The comment above this branch still states the old compile-time rule. On Arm32 and RV32 the PLT does arrange the `GOT[1]`/`GOT[2]` hand-off, and these entries are emitted because the user asked for `-z now`, not because lazy resolution is impossible. Reword it to say the binding mode is selected by `imm_binding` and forced on the targets whose PLT has no lazy path.

src/main.c
```diff
@@ -145,11 +156,22 @@ int main(int argc, char *argv[])
...
     }
 
+    if (dynlink) {
+        switch (ELF_MACHINE) {
+        /* The following 64-bit targets have no lazy-resolution path, so
+         * immediate binding must be used.
+         */
+        case ELF_MACHINE_X86_64:
+        case ELF_MACHINE_AARCH64:
+            imm_binding = true;
```
jserv, line 166: An explicit `-z lazy` is parsed, accepted, and then silently overridden here, so the user gets immediate binding with no diagnostic, unlike the other incompatible option pairs that `usage_error` rejects. `-z` without `--dynlink` is likewise accepted and does nothing. Recording whether `-z` was given, then rejecting `-z lazy` on these targets and `-z` without `--dynlink`, keeps the option contract honest.

src/main.c
```diff
@@ -145,11 +156,22 @@ int main(int argc, char *argv[])
...
+        }
+    }
+
     if (!in) {
         printf(
             "Usage: shecc [-I directory] [-o output] [+m] [--dot] [--dump-ir] "
             "[--warn-string-literals] [--std=c99] [--no-libc] "
-            "[--dynlink] [-E] <input.c>\n");
+            "[--dynlink] [-z <lazy | now>] [-E] <input.c>\n");
```
jserv, line 174: The usage string gains `-z`, but the usage line and option list in `README.md` still document the old set, and `docs/dynamic-linking.md` still lists "`DT_BIND_NOW` (force immediate binding) is not set" under Limitations, which is wrong for every target once this lands. Both belong in this change.

tests/driver.sh
```diff
@@ -106,7 +106,7 @@ case "$1" in
 esac
 
 if [ $# -ge 2 ] && [ "$2" = "1" ]; then
-    readonly SHECC_CFLAGS="--dynlink"
+    readonly SHECC_CFLAGS="--dynlink ${3:-lazy}"
```
jserv, line 109: `SHECC_CFLAGS` gets the binding mode without the `-z` that introduces it, so shecc sees a bare `lazy` or `now` operand. That operand is parsed as an input file and then overwritten by the real source that follows it, so `make check BINDING=now` runs the whole driver suite with lazy binding and reports success.

```suggestion
    readonly SHECC_CFLAGS="--dynlink -z ${3:-lazy}"
```

src/elf.c
DrXiao: @jserv I would like to ask about the x64 and Arm64 implementations before improving this pull request.

In the `x64.mk` and `arm64.mk`, `DYN_BIND_NOW` is always enabled, forcing immediate binding. However, if I understand correctly, both of these 64-bit architectures also support lazy binding. 

Since I wasn't involved in the review of the related pull requests, could you clarify why these targets are forced to use immediate binding?

DrXiao: > Since I wasn't involved in the review of the related pull requests, could you clarify why these targets are forced to use immediate binding?

By the way, the x64 and Arm64 are supported, but the documentation hasn't been updated to explain the implementation considerations. I can help update the documentation in a separate pull request.

jserv: > In the `x64.mk` and `arm64.mk`, `DYN_BIND_NOW` is always enabled, forcing immediate binding. However, if I understand correctly, both of these 64-bit architectures also support lazy binding.

I was standardizing the behavior across all backends. Feel free to consolidate the linkage logic as well.

PR #339, opened 2026-09-10
Test an escaped parameter read before its slot exists

### LINKED ISSUE #337: Taking the address of a parameter corrupts an expression computed before it
## Summary

`prepare_operand()` reloads an address-taken variable from its stack slot
rather than reading the register it is already in, since a store through the
pointer would not have reached the register. The reload needs the slot to hold
the value. A parameter arrives in a register and has no slot: one is reserved
the first time something spills it. Reloading before that reads whatever the
frame happened to hold.

Nothing has to be stored through the pointer. Taking the address is enough,
and the value is already wrong by the time the pointer exists.

## Reproduction

```c
#include <stdio.h>

int f(int a, int b)
{
    int x = a - b;
    int *p = &b;
    return x;
}

int main()
{
    printf("%d\n", f(5, 2));
    return 0;
}
```

| | result |
|---|---|
| `gcc -O0` | 3 |
| shecc x64 | **5** |
| shecc arm | **5** |
| shecc arm64 | **5** |
| shecc riscv | **5** |

5 is `a`: the subtraction read zero for `b`.

Tested at 4d3b4c0, on all four targets.

## Cause

`--dump-ir` for `f`:

```
f:
	load %x2, 0(sp)       ; b, from a slot nothing has written
	%x3 = sub %x0, %x2
	store %x2, 16(sp)     ; b given a slot only here
	%x0 = %sp + 16        ; p = &b
	ret %x3
```

`b` arrives in `x1`. The load that precedes the subtraction reads offset 0,
which no instruction has written; the store that gives `b` a home comes after,
and by then the wrong value has already been subtracted. The slot happened to
contain zero, which is why the result is `a` rather than something arbitrary.

The shape is narrow, which is probably why it has gone unnoticed: it needs the
escaped variable to be a parameter (a local is stored to its slot when it is
initialised), and the address to be taken after the expression rather than
before.

## Suggested fix

When the variable is in a register, is address-taken and has no slot yet,
write the register back before the reload and go on using that register. That
is what makes the slot stand for the variable, and it is what the code would
have done anyway the first time anything spilled it.

I have this working locally, with a regression test covering the subtraction,
the same shape around an addition, and the address taken of the first operand
rather than the second. The suite passes on all four targets at both stages.
Happy to send a pull request.

One caveat I am aware of: this is in `prepare_operand()`, which everything
goes through, so I would rather someone who knows the allocator confirmed the
placement is right before it lands. A narrower fix in whatever gives
parameters their slots might be more appropriate -- I could not find a natural
place to do it there, because the slot is deliberately allocated lazily.

alanhc:
Rebased on `2d38107`. The rebase conflicted in `src/reg-alloc.c`, and the
conflict turned out to be the point: master already fixes #337.

`pin_registers()`'s caller marks an incoming argument register `polluted = 1`
on function entry, so the `|| REGS[i].polluted` arm of `prepare_operand()`
returns the register instead of reloading it, and the reload that read a slot
the parameter never had is no longer reached. The comment above that condition
already names this case.

Comparing what stage 0 emits for `sub_then_escape()`:

    4d3b4c0 (this branch's old base)     master
    load %x2, 0(sp)                      %x2 = sub %x0, %x1
    %x3 = sub %x0, %x2                   store %x1, 16(sp)
    store %x2, 16(sp)                    %x0 = %sp + 16
    %x0 = %sp + 16                       ret %x2
    ret %x3

The three functions in the test all come out right on master, so the
`address_taken && !space_is_allocated` hunk this PR carried is dead code now.
Dropped it.

What is left is the regression test: `tests/escaped-param.c` covers the
subtraction, the same shape around an addition, and the address taken of the
first operand rather than the second. The shape is narrow -- it needs the
escaped variable to be a parameter, and the address taken after the expression
rather than before -- which is why it went uncovered until it broke.

Close this instead if the test is not worth a file of its own.

Regression test for #337, which master already fixes.


jserv (changes requested):
Rebase latest `master` branch and resolve conflicts.

alanhc:
Rebased on `2d38107`. The conflict was in `src/reg-alloc.c`, and it is the
answer: master already fixes #337. An incoming argument register is marked
`polluted = 1` on function entry, so `prepare_operand()` takes the
`|| REGS[i].polluted` arm and returns the register instead of reloading it
from a slot the parameter does not have yet.

Stage 0 for `sub_then_escape()`, at this branch's old base `4d3b4c0` versus
master:

    load %x2, 0(sp)          %x2 = sub %x0, %x1
    %x3 = sub %x0, %x2       store %x1, 16(sp)
    store %x2, 16(sp)        %x0 = %sp + 16
    %x0 = %sp + 16           ret %x2
    ret %x3

All three functions in the test come out right on master, so the
`address_taken && !space_is_allocated` hunk is dead code and is dropped. What
remains is `tests/escaped-param.c`. Close this if the test is not worth its own
file.

tests/escaped-param.c
```diff
@@ -0,0 +1,41 @@
+/*
```
jserv, line 1: The file opens with an empty line where its sibling `tests/strength-reduce.c` opens with a comment naming the shape it guards. This test depends on a narrow shape (a parameter still sitting in its arrival register, its address taken only after the expression reads it), and nothing in the file says so. A later cleanup could drop the unused `p` or move `&b` above the arithmetic, and the test would still pass while covering nothing. Move the explanation from the pull request description into the file:

```suggestion
/* A parameter whose address is taken only after an expression has
 * read it is still in the register it arrived in, and its stack slot may not
 * exist yet. Reloading it from that slot read whatever sat at offset
 * zero of the frame. Keep the `&` after the arithmetic and the
 * pointers in place: both are what make this path reachable.
 */
```

alanhc: Applied. The file has carried a header comment since the first push, but it explained why no store through the pointer is needed rather than warning against moving the `&` or dropping `p`, which is the part a later cleanup would break — your wording covers that, so I took it as is.

One deviation: `.ci/check-commentflow.sh` reflows the opening `/*` onto its own line, matching `tests/strength-reduce.c`. The text is yours; the layout is commentflow's output.

jserv:
Thank @alanhc for contributing!

